PR #23, opened 2022-01-11
rename github token

nenadjaja:


.github/workflows/ci-cd-pull-request.yml
```diff
@@ -20,7 +20,7 @@ jobs:
         with:
           registry: ghcr.io
           username: Jannis
```
benface, line 22: Do we have to update this as well @nenadjaja?

nenadjaja: good question. I should update it to my account

benface: Should it be another "secret" (environment variable), since it might change?

benface (approved):

PR #38, opened 2022-01-21
Fix formatting in indexer FAQ infrastructure list

drklee3:
* Removed extra space in bold items which breaks formatting
* Made Prometheus to actually be it's own item instead of at the end of "Indexer agent"
* Removed duplicated "PostgreSQL"s


azf20 (approved):
Thanks, LGTM!

PR #39, opened 2022-01-23
Update quick-start.mdx

skundu42:
Update Matic to Polygon to prevent confusion


azf20 (commented):
Thanks @skundu42! While making that change, does "Binance" make sense here? Perhaps we could mention a rollup (Optimism or Arbitrum) to clarify that this is also the case for rollups (as well as other chains). WDYT @schmidsi

azf20:
Closing as quick start has changed, thanks @skundu42 and sorry for the slow response here

PR #40, opened 2022-01-23
network name update

skundu42:
Update Matic to Polygon to prevent confusion and maintain consistency.

### DIFF
--- a/pages/en/developer/create-subgraph-hosted.mdx
+++ b/pages/en/developer/create-subgraph-hosted.mdx
@@ -24,7 +24,7 @@ The Graph Network supports subgraphs indexing mainnet Ethereum:
 - `xdai`
 - `near-mainnet`
 - `near-testnet`
-- `matic`
+- `polygon`
 - `mumbai`
 - `fantom`
 - `bsc`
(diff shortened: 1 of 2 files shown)


pages/en/developer/create-subgraph-hosted.mdx
```diff
@@ -24,7 +24,7 @@ The Graph Network supports subgraphs indexing mainnet Ethereum:
 - `xdai`
 - `near-mainnet`
 - `near-testnet`
-- `matic`
```
azf20, line 27: This naming is currently fixed in the Hosted Service. There is an [open issue](https://github.com/graphprotocol/graph-node/issues/3075) that will enable easy switches such as this, but we can't change this naming at this time (we can perhaps add a note to indicate that this is "Polygon" chain)

azf20 (changes requested):

azf20:
Thanks @skundu42 ! We can't fully switch over at this time (blocked by this [issue](https://github.com/graphprotocol/graph-node/issues/3075)), but we can perhaps add some clarifiers. What do you think @schmidsi ?

PR #61, opened 2022-02-14
doc(developer): complete beginner guide to querying a GraphQL API

charlypoly:
Hi, here Charly from The Guild 

@azf20, as discussed together, here is the new content for the "Querying from an Application" documentation page

Let me know how to proceed 👀 
Also, how will be handled the existing translated version of the old "Querying from an Application" page?

### DIFF
--- a/pages/en/developer/querying-from-your-app.mdx
+++ b/pages/en/developer/querying-from-your-app.mdx
@@ -239,29 +239,30 @@ Note: The opposite directive is `@skip(if: ...)`.
 
 **"Ask for what you want"**
 
-GraphQL became famous for its “Ask for what you want” tagline.
+GraphQL became famous for its "Ask for what you want" tagline.
 
 For this reason, there is no way, in GraphQL, to get all available fields without having to list them individually.
 
 When querying GraphQL APIs, always think of querying only the fields that will be actually used.
 
-Since some fields might be computed or proxied on the server-side, fetching fewer fields can make a big difference!
+A common cause of over-fetching is collections of entities. By default, queries will fetch 100 entities in a collection, which is usually much more than what will actually be used, e.g., for display to the user. Queries should therefore almost always set first explicitly, and make sure they only fetch as many entities as they actually need. This applies not just to top-level collections in a query, but even more so to nested collections of entities.
+
+For example, in the following query:
 
 ```graphql
-query getProtocols {
-  protocols(first: 5) {
+query listTokens {
+  tokens { # will fetch up to 100 tokens
     id
-    inflation
-    inflationChange
-    maxEarningsClaimsRounds
-    numActiveTranscoders
+    transactions { # will fetch up to 100 transactions
+      id
+    }
   }
 }
 ```
 
-`numActiveTranscoders` might be slower (computed) field to resolve.
+The response could contain 100 transactions for each of the 100 tokens.
 
-If not needed, removing it will improve the overall query performance (*especially in a list context:* `list: 5`).
+If the application only needs 10 transactions, the query should explicitly set `first: 10` on the transactions field.
 
 **Combining multiple queries**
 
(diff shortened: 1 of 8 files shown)


benface:
Hi @charlypoly! About your translations question, I think the right way to proceed would be to overwrite the file in all languages with your changes (so basically duplicate your `querying-from-your-app.mdx` file in each `pages/[language]/developer` folder). When this PR is merged, Crowdin will pick up the new source strings so translators will be able to translate them, and subsequently Crowdin will propose changes to the files in the different languages (except English) via a pull request. Does that make sense?

pages/en/developer/querying-from-your-app.mdx
```diff
@@ -2,135 +2,572 @@
...
+    maxEarningsClaimsRounds
+    numActiveTranscoders
+  }
+}
 ```
 
-Then you can query the API with the following code:
+`numActiveTranscoders` might be slower (computed) field to resolve.
+
+If not needed, removing it will improve the overall query performance (*especially in a list context:* `list: 5`).
```
lutter, line 264: I think that should say `first: 5`.

pages/en/developer/querying-from-your-app.mdx
```diff
@@ -2,135 +2,572 @@
...
-First, install `urql` and `graphql`:
+GraphQL became famous for its “Ask for what you want” tagline.
 
-```sh
-npm install urql graphql
+For this reason, there is no way, in GraphQL, to get all available fields without having to list them individually.
+
+When querying GraphQL APIs, always think of querying only the fields that will be actually used.
+
+Since some fields might be computed or proxied on the server-side, fetching fewer fields can make a big difference!
```
lutter, line 248: We don't have computed fields (yet) A much bigger concern is overfetching of collections since the default value for `first` is `100`. It would be good to add a paragraph about that, something like:

A common cause of overfetching are collections of entities. By default, queries will fetch 100 entities in a collection, which is usually much more than what will actually be used, e.g., for display to the user. Queries should therefore almost always set `first` explicitly, and make sure they only fetch as many entities as they actually need. This applies not just to top-level collections in a query, but even more so to nested collections of entities. For example, in the query
```graphql
query listTokens {
  tokens {
    id
    transactions { id }
  }
}
```
the response could contain 100 transactions for each of the 100 tokens in the response. If the application only needs 10 transactions, the query should set `first: 10` explicitly on the `transactions` field.

pages/en/developer/querying-from-your-app.mdx
```diff
@@ -2,135 +2,572 @@
...
+  poll {
+    id
+    proposal
+  }
+}
+```
+
+---
+
+## The essential tools
```
lutter, line 459: Nice! I would add some browser based tools, like [Altair](https://altair.sirmuel.design/) or [GraphiQL](https://graphiql-online.com/graphiql) since they are a very low-overhead way to check queries for problems.

dotansimha (approved):

pages/en/developer/querying-from-your-app.mdx
charlypoly: Done in b656828

pages/en/developer/querying-from-your-app.mdx
charlypoly: Thanks for the review @lutter!

I added your suggestion via b656828.

lutter (approved):
Nice! Thanks for the additions

azf20 (approved):
Thanks @charlypoly LGTM!

azf20:
@benface as discussed this has been moved to be its own separate page, so we can get it merged and translated (and then we can migrate / show it in the navbar). Hope that makes sense! Will hold off merging to check we have done this right 🙏

benface:
This is great @charlypoly and @azf20, thank you!

charlypoly:
Hi @azf20!

Any news on this PR? 👀

azf20:
Hey @charlypoly sorry was away these past weeks - back now and merged it, hopefully can get translations up and running and can go into the next release!

charlypoly:
Thank you @azf20!

PR #66, opened 2022-02-18
IDs can be bytes as well as strings

azf20:


evaporei (approved):

azf20:
more detailed fix here https://github.com/graphprotocol/docs/pull/106

PR #75, opened 2022-03-03
Move i18n module to `components`

benface:


IainM22 (approved):
👌

layout/MDXLayoutPagination.tsx
```diff
@@ -29,8 +29,8 @@ export const MDXLayoutPagination = () => {
             >
               <Flex.Column align="start" gap={Spacing.S} sx={{ textAlign: 'left' }}>
                 <Flex.Row align="center" gap={Spacing.S}>
-                  <Icon.ArrowLeft size="12px" />
-                  <Text.S10>{translations.global.previous}</Text.S10>
+                  <Icon.ArrowLeft title="" size="12px" />
```
dnfodjo, line 32: is this empty text necessary

layout/MDXLayoutPagination.tsx
```diff
@@ -67,8 +67,8 @@ export const MDXLayoutPagination = () => {
             >
               <Flex.Column align="end" gap={Spacing.S} sx={{ textAlign: 'right' }}>
                 <Flex.Row align="center" gap={Spacing.S}>
-                  <Text.S10>{translations.global.next}</Text.S10>
-                  <Icon.ArrowRight size="12px" />
+                  <Text.S10>{t('global.next')}</Text.S10>
+                  <Icon.ArrowRight title="" size="12px" />
```
dnfodjo, line 71: same here

cmwhited (approved):
Without having much insight into the translations process, from what I can tell, this looks good. Nothing sticks out. Good work.

layout/MDXLayoutPagination.tsx
benface: Yes, because all icons have a default title, but in this case we don't want it to be read by screen readers. The button already has accessible text.

benface: This is the equivalent of doing `alt=""` on an image to mark it as decorative.

dnfodjo (approved):
looks good. Great job

PR #86, opened 2022-03-10
Update matchstick examples

cargopete:
Updating [Matchstick ](https://github.com/LimeChain/matchstick) examples with the latest version of the framework.

### DIFF
--- a/pages/en/developer/matchstick.mdx
+++ b/pages/en/developer/matchstick.mdx
@@ -61,7 +61,7 @@ export function createNewGravatarEvent(
 }
 ```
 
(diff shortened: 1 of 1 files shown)


PtomelySoter:
This is looking great! Do we have examples for setting up tests for callHandlers? It would also be good to add in that event parameters need to match the order of the event itself.

PtomelySoter:
Happy to come up with some copy for this. @axiomatic-aardvark

schmidsi (approved):
LGTM. Great improvements.

azf20 (approved):
Great stuff!

azf20:
Thanks @axiomatic-aardvark! @PtomelySoter definitely good idea, but I think best to address in a separate PR

PR #87, opened 2022-03-14
Fixing typo in conversion function BigInt

juanigallo:
In the conversion table, the row

`int32 | BigInt | Bigint.fromI32(s)`

should be:

`int32 | BigInt | BigInt.fromI32(s)`


azf20 (approved):
LGTM. Good spot!

PR #88, opened 2022-03-16
Add `Layout` component with sticky header

alexandriaroberts:
Importing Layout component here. I just did minimal changed to _app.js file for the background image in div.

Issue: https://app.asana.com/0/1200559321875810/1201921576460054/f


pages/_app.tsx
```diff
@@ -75,8 +87,32 @@ const MyApp = ({ Component, pageProps }: AppProps) => {
...
+          />
+          <Layout
+            headerSticky
+            headerContent={
+              <NavigationMarketing activeRoute="/docs" NextLink={NextLink} rightAlignItems={[localeSwitcher]} />
+            }
+            mainContainer
+            footerContent={<Footer localeSwitcher={localeSwitcher} />}
+          >
+            <div>
```
benface, line 113: Hey @alexandriaroberts do you remember why you wrapped `<Component {...pageProps} />` in a `<div>` here? It doesn't seem required, but maybe I'm missing something.

benface (approved):

pages/_app.tsx
alexandriaroberts: To be honest, I don't remember now but I think it was left there by mistake. It isn't necessary there. Thank you very much for pointing it out. Removing it. 🙂

PR #99, opened 2022-03-24
[update] Reverse Lookups. Many-to-many examples

ubinatus:
With latest `@graphprotocol/graph-cli@0.28.1` and `@graphprotocol/graph-ts@0.26.0`, GraphQL schemas can't have Lists which nullable members. 
Basically solves by changing `[UserOrganization]!` to `[UserOrganization!]!`


azf20 (approved):
LGTM, thanks!

PR #101, opened 2022-03-30
Update matchstick docs

dimitrovmaksim:
In order for all the needed info to be in one place, I have added all the info from the matchstick repo as a new `Getting Started` section in the matchstick docs. Now users don't need to follow links to the repo or demo-subgraph to see how to run docker, what dependencies should be installed, or how to configure custom test and/or libs folders.
Also adds instructions for mocking IPFS files and how to run matchstick in coverage mode.

@graphprotocol/docs-reviewers

### DIFF
--- a/pages/en/developer/matchstick.mdx
+++ b/pages/en/developer/matchstick.mdx
@@ -4,8 +4,6 @@ title: Unit Testing Framework
 
 Matchstick is a unit testing framework, developed by [LimeChain](https://limechain.tech/), that enables subgraph developers to test their mapping logic in a sandboxed environment and deploy their subgraphs with confidence!
 
-Follow the [Matchstick installation guide](https://github.com/LimeChain/matchstick/blob/main/README.md#quick-start-) to install. Now, you can move on to writing your first unit test.
-
 ## Getting Started
 
 ### Install dependencies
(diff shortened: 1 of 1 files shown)


GundamDweeb:
Have reviewed - Looks good!

pages/en/developer/matchstick.mdx
```diff
@@ -196,6 +249,94 @@ createMockedFunction(contractAddress, 'getGravatar', 'getGravatar(address):(stri
...
+NOTE: When testing `ipfs.map/ipfs.mapJSON`, the callback function must be exported from the test file in order for matchstck to detect it, like the `processGravatar()` function in the test example bellow:
+
+`.test.ts` file:
+
+```typescript
+import { assert, test, mockIpfsFile } from "matchstick-as/assembly/index"
+import { ipfs } from "@graphprotocol/graph-ts"
+import { gravatarFromIpfs } from "./utils"
+
+# Export ipfs.map() callback in order for matchstck to detect it
```
schmidsi, line 264: This is not a proper comment in TypeScript/AssemblyScript

pages/en/developer/matchstick.mdx
```diff
@@ -4,15 +4,68 @@ title: Unit Testing Framework
...
+## Getting Started
+
+### Install dependencies
+In order to use the test helper methods and run the tests, you will need to install the following dependencies:
+
+```sh
+yarn add --dev matchstick-as
+```
+
+❗ Since `graph-node` depends on diesel (and diesel requires a local postgres installation) we highly advise using the commands below as adding it in any other way may cause unexpected errors! If you don't have Postgres installed, you will need to install it:
```
schmidsi, line 16: The dependency on diesel does not really help here and is just confusing in my opinion. It is sufficient to say that graph-node depends on Postgres.

schmidsi (changes requested):
Thanks a lot for this submission. This makes a lot of sense. I just added some small corrections, otherwise we are good to go.

schmidsi (approved):

PR #104, opened 2022-04-06
Make debug forking docs stable :)

evaporei:
- Adds debug-forking to navigation bar again
- Removes v0.26.0 warning since it's out!

Reference: https://github.com/graphprotocol/graph-node/blob/master/NEWS.md#0260


azf20 (approved):
This LGTM!

PR #106, opened 2022-04-13
Document `id: Bytes!` and immutable entities

lutter:
* remove all mentions of the `ID` type from the docs and urge users to use `id: Bytes`
* explain the `@entity(immutable: true)` annotation

### DIFF
--- a/pages/en/developer/create-subgraph-hosted.mdx
+++ b/pages/en/developer/create-subgraph-hosted.mdx
@@ -166,14 +166,14 @@ The schema for your subgraph is in the file `schema.graphql`. GraphQL schemas ar
 
 Before defining entities, it is important to take a step back and think about how your data is structured and linked. All queries will be made against the data model defined in the subgraph schema and the entities indexed by the subgraph. Because of this, it is good to define the subgraph schema in a way that matches the needs of your dapp. It may be useful to imagine entities as "objects containing data", rather than as events or functions.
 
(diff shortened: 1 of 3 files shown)


pages/en/developer/create-subgraph-hosted.mdx
```diff
@@ -210,7 +210,9 @@ Entity fields can be defined as required or optional. Required fields are indica
 Null value resolved for non-null field 'name'
 ```
 
-Each entity must have an `id` field, which is of type `ID!` (string). The `id` field serves as the primary key, and needs to be unique among all entities of the same type.
+Each entity must have an `id` field, which must be of type `Bytes!` or `String!`. It is generally recommended to use `Bytes!`, unless the `id` contains human-readable text, since entities with `Bytes!` id's will be faster to write and query as those with a `String!` `id`. The `id` field serves as the primary key, and needs to be unique among all entities of the same type. For historical reasons, the type `ID!` is also accepted and is a synonym for `String!`.
```
schmidsi, line 213: I think this is a typo:
```
`String!` `id`. The `id`
```
Should be:
```
`String!`. The `id`
```

schmidsi (changes requested):
Just one very small typo

pages/en/developer/create-subgraph-hosted.mdx
lutter: No, I meant to say 'string id' here. All the backquotes make it a little hard to see, but if I read the sentence out loud, it makes sense to me as written.

schmidsi (approved):

PR #110, opened 2022-04-21
Corrected entry name for blockHandlers

tnkrxyz:
blockHandlers do not have function entry, but only handler


azf20 (approved):
Good catch, thanks!

tnkrxyz:
This correction needs to be propagated to translation in other languages. Since I don't understand any of those, I didn't mess with them; somebody has to do it.

azf20:
@tnkrxyz yep all good, that'll get triggered when merged. Thanks!

PR #111, opened 2022-04-22
Typo: bnb -> bsc, add explanatory ntoes

azf20:
- Add explanatory notes for Polygon, Gnosis Chain and BNB Chain

### DIFF
--- a/pages/en/developer/create-subgraph-hosted.mdx
+++ b/pages/en/developer/create-subgraph-hosted.mdx
@@ -27,7 +27,7 @@ The Graph Network supports subgraphs indexing mainnet Ethereum:
 - `matic`
 - `mumbai`
 - `fantom`
-- `bsc`
+- `bsc` (now known as BNB chain)
 - `chapel`
 - `clover`
 - `avalanche`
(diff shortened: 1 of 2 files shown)


pages/en/developer/create-subgraph-hosted.mdx
```diff
@@ -27,7 +27,7 @@ The Graph Network supports subgraphs indexing mainnet Ethereum:
 - `matic`
 - `mumbai`
 - `fantom`
-- `bnb`
+- `bsc`
```
maoueh, line 30: Although the network identifier needs to still be `bsc`, the chain was indeed renamed to `BNB`, do we want to add this fact in there, maybe like suggested edit.
```suggestion
- `bsc` (now known as BNB chain)
```

Or maybe we are able to update deployment to `bnb` and have an alias?

this-username-is-taken: I guess the same case can be made for xdai and matic. But a comment can definitely helpful, especially for people who entered the space recently

azf20:
Good ideas, updated accordingly

benface (approved):

PR #117, opened 2022-05-02
Added docs for new input filters available in `graph-node`

dotansimha:
Documentation for:
- https://github.com/graphprotocol/graph-node/pull/3294 
- https://github.com/graphprotocol/graph-node/pull/3186

### DIFF
--- a/pages/en/developer/graphql-api.mdx
+++ b/pages/en/developer/graphql-api.mdx
@@ -57,7 +57,7 @@ Further, the `skip` parameter can be used to skip entities and paginate. e.g. `f
 
 Queries should avoid using very large `skip` values since they generally perform poorly. For retrieving a large number of items, it is much better to page through entities based on an attribute as shown in the last example.
 
-#### Example
+#### Example using `first`
 
 Query the first 10 tokens:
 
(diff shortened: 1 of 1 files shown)


saihaj (approved):

pages/en/developer/graphql-api.mdx
```diff
@@ -134,9 +134,25 @@ You can use suffixes like `_gt`, `_lte` for value comparison:
 }
 ```
 
-Full list of parameter suffixes:
+#### Example for block filtering
+
+You can also filter entities by the `_change_block`:
```
azf20, line 139: Do you think you could add a little more information here on what this means, and what it enables?
i.e. This filters entities which were updated in or after the specified block. This can be useful if you are looking to fetch only entities which have changed, for example since the last time you polled. Or alternatively it can be useful to investigate or debug how entities are changing in your subgraph (if combined with a block filter, you can isolate only entities that changed in a specific block)

pages/en/developer/graphql-api.mdx
```diff
@@ -145,11 +161,23 @@ _lte
...
 _not_starts_with
+_not_starts_with_nocase
 _not_ends_with
+_not_ends_with_nocase
+```
+
+In addition, the following global filters are available as part of `where` argument:
+
+```gr
+_change_block
```
azf20, line 180: is it worth mentioning that there is only one supported filter here (`number_gte`)?

azf20 (changes requested):

azf20:
Thanks @dotansimha! Would be great to have just a little more context, particularly on `number_gte`. Happy to discuss!

pages/en/developer/graphql-api.mdx
dotansimha: sure, done

azf20 (approved):
LGTM thanks!

PR #129, opened 2022-05-15
Adding troubleshooting tips for Matchstick on WSL.

cargopete:
I added some tips/workarounds for common issues that users have faced recently with Matchstick, especially on WSL.


axiomatic-aardvark:
Closing since the Docker workaround turned out to be redundant and the other tips are pinned in Discord + added to the Matchstick docs, I don't think they're important enough to clutter the graph docs too.

PR #147, opened 2022-06-10
Add Moonbeam/Moonriver to list of supported networks (docs)

melotik:
- It seems like the docs were never updated from when `moonbeam` was added to the list of networks supported on the Hosted Service.

- Also, `Moonriver` was missing from the list of supported networks in the developer FAQs.

This fixes that in every language.


dmelotik:
this is a repeat PR, closing

PR #149, opened 2022-06-15
Update dependencies + run Prettier

benface:
This removes the "Graph Day" and "Graph Hack" pills from the header, and adds a "Past Events" dropdown to the footer.

CleanShot 2022-06-15 at 15 07 27@2x

CleanShot 2022-06-15 at 15 14 32@2x


crowdin.yml
```diff
@@ -4,9 +4,7 @@ files:
     translation: /pages/%two_letters_code%/**/%original_file_name%
     dest: /**/%original_file_name%
     content_segmentation: 0
-    skip_untranslated_files: true
```
benface, line 7: Turns out these options are not supported by Crowdin's GitHub integration, so removing. I enabled them at the project level instead (we can't get file type granularity that way, but that's OK).

IainM22 (approved):

PR #156, opened 2022-06-23
Consolidate Supported Networks list

ahmadmardeni1:


schmidsi (changes requested):
Hmm, supported networks only applies to the hosted service & decentralised network. So it belongs under “Publishing” in my opinion. One can basically run their own graph-node indexing whatever EVM compatible chain.

ahmadmardeni1:
From a dev perspective, I want to know if the network I’m building a subgraph to index is supported in the first place before “creating the subgraph”.
What do you think? @schmidsi

pages/en/developing/supported-networks.mdx
```diff
@@ -0,0 +1,53 @@
...
+## The Graph Decentralized Network
+
+The Graph Network supports subgraphs indexing the Ethereum mainnet and Rinkeby testnet:
+
+- `mainnet`
+- `rinkeby`
+
+## Hosted Service
+
+**The following networks are supported in beta on the Hosted Service**:
```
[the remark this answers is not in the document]
ahmadmardeni1: I copy-pasted from what we currently have, but sure I will change it!

ahmadmardeni1:
Done! @benface

benface (approved):

azf20:
I think I agree with Simon's point. I think the key goal here is to help developers develop an accurate mental model

Under "Developing", I think it's important to convey which protocols are supported.
- Graph Node can index any EVM compatible network.
- Graph Node can also index other protocols, via a Firehose integration. Firehose integrations have been created for NEAR, Arweave and Cosmos-based chains

The Hosted Service's Graph Nodes can index networks across all supported protocols. See the list of specific networks supported here -> linking to a "Deploying on the Hosted service" page (and I think it makes sense to split the list of supported by protocol, rather than mixing in NEAR with EVM etc.)

Indexers on The Graph Network can offer full support & security for mainnet Ethereum subgraphs. Please see more information on deploying to the network here -> linking to Deploying to the Studio page, where the key is to convey information about what is supported (mainnet) and why (proof of indexing & determinism)

So it's not quite consolidating into one place, but I think is an accurate depiction

schmidsi:
I think @ahmadmardeni1 point is valid that in a developer journey to The Graph, one might want to find out about the supported networks before dedicating more time to the technology. In my opinion, this should be part of the marketing that happens before the docs. We are reworking the landing page for this.

Otherwise, I agree with Adam. It would also incentivise developers to use the graph-node on chains that are not supported on the hosted service.

pages/en/developing/supported-networks.mdx
```diff
@@ -4,52 +4,27 @@ title: Supported Networks
...
+- Near
+- Cosmos
+- Arweave 
+
+See the list of which networks are supported [here](deploying/deploying-a-subgraph-to-hosted).
 
 > The Graph's Hosted Service relies on the stability and reliability of the underlying technologies, namely the provided JSON RPC endpoints. Newer networks will be marked as being in beta until the network has proven itself in terms of stability, reliability, and scalability. During this beta period, there is a risk of downtime and unexpected behaviour.
 
-Remember that you will **not be able** to publish a subgraph that indexes a non-mainnet network to the decentralized Graph Network in [Subgraph Studio](/deploying/subgraph-studio).
+> Graph Node can index any EVM compatible chain.
```
azf20, line 28: I think worth moving these notes to the top, and adding "Specific support depends where you are deploying your subgraph".

azf20 (approved):
Looking good! One small suggestion to move the description of support to the top of the page

pages/en/developing/supported-networks.mdx
ahmadmardeni1: Take a look! Hopefully, this is what you meant :) @azf20

pages/en/developing/supported-networks.mdx
```diff
@@ -12,6 +12,12 @@ Indexers on The Graph Network can offer full support & security for mainnet Ethe
...
 
+## Graph Node
+
+You can run your own Graph Node to index any EVM-compatible chain. Make sure that the [version](https://github.com/graphprotocol/graph-node/releases) you are using supports the network and you have the needed configuration.
+
+Graph Node can also index other protocols, via a Firehose integration. Firehose integrations have been created for NEAR, Arweave and Cosmos-based chains.
+
 ## Hosted Service
 
 The following protocols are supported on the Hosted Service:
```
schmidsi, line 23: This sentence does not make sense anymore.

schmidsi (changes requested):

schmidsi (approved):

PR #161, opened 2022-06-30
New structure

benface:
### Current (old) structure

CleanShot 2022-06-30 at 12 35 39@2x

### Proposed (new) structure

CleanShot 2022-06-30 at 12 38 23@2x


azf20 (approved):
Thanks so much Benoit!

PR #162, opened 2022-07-04
Add that now `deploy` also accepts the `--network` and `--network-file` options

dimitrovmaksim:


pages/en/hosted-service/deploy-subgraph-hosted.mdx
```diff
@@ -52,16 +52,14 @@ In some cases, you will want to deploy the same subgraph to multiple Ethereum ne
 
 ### graph-cli >=0.29.0
```
benface, line 53: Should we change this heading since the options were only added in `v0.32.0` for `graph deploy`?

dimitrovmaksim: I wasn't sure about that either, because the options are available in graph-cli for the `build` command since v0.29.0, we just made them available for `deploy` too, and now you can use whichever.

dimitrovmaksim: Or I can change the heading to `Using graph-cli` and `Using templates/templating` or something of that sort

benface: > Or I can change the heading to Using graph-cli and Using templates/templating or something of that sort

I think that would be better! 👍

benface (approved):
This is great, thank you!

dimitrovmaksim:
> This is great, thank you!

Hey, I fixed a merge conflict caused by the new structure PR, fixed a typo and added a Note about the `deploy` command at the end of the example

benface (approved):
🙌

PR #163, opened 2022-07-06
fixes typos

i-am-sam-codes:
line 86: directly follow -> directly followed 
line 142: removes "it"
line 175: stored -> store
line 191: is -> are - adds comma before etc


pages/en/developer/assemblyscript-api.mdx
```diff
@@ -188,7 +188,7 @@ The `TypedMap` class has the following API:
 import { Bytes } from '@graphprotocol/graph-ts'
 ```
 
-`Bytes` is used to represent arbitrary-length arrays of bytes. This includes Ethereum values of type `bytes`, `bytes32` etc.
+`Bytes` are used to represent arbitrary-length arrays of bytes. This includes Ethereum values of type `bytes`, `bytes32`, etc.
```
benface, line 191: I think this was in the sense "[The class] `Bytes` is used to represent..."

So, I would leave it as is:
```suggestion
`Bytes` is used to represent arbitrary-length arrays of bytes. This includes Ethereum values of type `bytes`, `bytes32`, etc.
```

benface (changes requested):
Thank you very much for your contribution! It all looks great other than the little thing I mentioned above.

benface (approved):
🙏

PR #176, opened 2022-07-26
Combine created and defined

azf20:
- Combining these two pages which sound very similar
- Including a redirect


nginx.conf
```diff
@@ -83,6 +83,7 @@ http {
         rewrite ^/docs/([a-zA-Z][a-zA-Z])/supported-networks/arweave/$                  $scheme://$http_host/docs/$1/cookbook/arweave/                              permanent;
         rewrite ^/docs/([a-zA-Z][a-zA-Z])/supported-networks/cosmos/$                   $scheme://$http_host/docs/$1/cookbook/cosmos/                               permanent;
         rewrite ^/docs/([a-zA-Z][a-zA-Z])/supported-networks/near/$                     $scheme://$http_host/docs/$1/cookbook/near/                                 permanent;
+        rewrite ^/docs/([a-zA-Z][a-zA-Z])/developing/defining-a-subgraph/$              $scheme://$http_host/docs/$1/developing/creating-a-subgraph/
```
benface, line 86: I notice this rule is the only one not to have the `permanent` flag, not sure if it's an oversight or if you think these redirects (or this one in particular) should _not_ be permanent?

benface (approved):
🙌

nginx.conf
azf20: That wasn't intentional, fixed, thanks!

PR #179, opened 2022-08-03
 Fix for arabic translations

alexandriaroberts:


ahmadmardeni1 (approved):

pages/_app.tsx
```diff
@@ -122,18 +123,18 @@ const MyApp = ({ Component, pageProps }: AppProps) => {
...
               minHeight: '768px',
+              aspectRatio: '1',
               backgroundImage: `url('${process.env.BASE_PATH}/img/page-background.png')`,
               backgroundSize: 'cover',
               backgroundPosition: 'center top',
-              '@media (min-width: 1440px)': {
-                aspectRatio: '1440/768',
-              },
             }}
           />
```
benface, line 131: This changes what the page looks like:

**Before**
CleanShot 2022-08-05 at 17 55 39@2x

**After**
CleanShot 2022-08-05 at 17 56 39@2x

I think a better solution would be to wrap this `div` in another position-absolute `div` that has `overflow: hidden`.

benface (changes requested):
Looks good, just the one comment I left + this little bug that I found on mobile (missing padding):

CleanShot 2022-08-05 at 18 00 10@2x

pages/_app.tsx
alexandriaroberts: Yes, that's true. That does work well, I have updated this. Thank you. :)

alexandriaroberts:
> Looks good, just the one comment I left + this little bug that I found on mobile (missing padding):
> 
> CleanShot 2022-08-05 at 18 00 10@2x

Thank you, glad this is coming a long. That's good fine, thank you! I have updated it. I will also send Ahmad the latest staging and ask for his opinion with this. :)

pages/_app.tsx
```diff
@@ -115,25 +116,27 @@ const MyApp = ({ Component, pageProps }: AppProps) => {
...
-              position: 'absolute',
-              left: 0,
-              right: 0,
-              top: 0,
-              minHeight: '768px',
-              backgroundImage: `url('${process.env.BASE_PATH}/img/page-background.png')`,
-              backgroundSize: 'cover',
-              backgroundPosition: 'center top',
-              '@media (min-width: 1440px)': {
+          <div sx={{ position: 'absolute', left: 0, right: 0, top: 0, bottom: 0, overflow: 'hidden' }}>
```
benface, line 119:
```suggestion
          <div sx={{ position: 'absolute', inset: 0, overflow: 'hidden' }}>
```

pages/_app.tsx
```diff
@@ -115,25 +116,27 @@ const MyApp = ({ Component, pageProps }: AppProps) => {
...
-            sx={{
-              position: 'absolute',
-              left: 0,
-              right: 0,
-              top: 0,
-              minHeight: '768px',
-              backgroundImage: `url('${process.env.BASE_PATH}/img/page-background.png')`,
-              backgroundSize: 'cover',
-              backgroundPosition: 'center top',
-              '@media (min-width: 1440px)': {
```
benface, line 128: We want to keep the media query for `aspectRatio`, so that on viewports smaller than 1440px, this `div` (in which the background image is centered) is 100% of the viewport's width (due to the `left: 0` and `right: 0`) and exactly 768px tall (due to the `minHeight`). But on viewports >= 1440px, the `div` keeps growing in height to maintain the `1440/768` aspect ratio, which is the same as the background image.

benface (commented):
Thanks for fixing the padding issue!

There's one more thing, sorry for not noticing before... this is kinda nitpicky, but I wonder if the diamond next to the active nav item should be on the right end side, to keep the logic consistent with `ltr` languages. Notice how there's padding on the left of the nav in English to make space for the diamonds, and that padding is also on the left in Arabic, but perhaps it should be on the right? I think this would be especially beneficial on mobile, where the diamond currently ends up pretty far from the text. We could do this in a separate PR, let me know what you think.

CleanShot 2022-08-09 at 12 04 35@2x

CleanShot 2022-08-09 at 12 02 46@2x

pages/_app.tsx
alexandriaroberts: Perfect, thank you. Have updated this. :)

pages/_app.tsx
alexandriaroberts: I removed it because there was an issue with the horizontal scroll and this is what it was causing it. But seems like it is fine now and it's not longer happening with media query. so I am adding it back, thank you!

alexandriaroberts:
> Thanks for fixing the padding issue!
> 
> There's one more thing, sorry for not noticing before... this is kinda nitpicky, but I wonder if the diamond next to the active nav item should be on the right end side, to keep the logic consistent with `ltr` languages. Notice how there's padding on the left of the nav in English to make space for the diamonds, and that padding is also on the left in Arabic, but perhaps it should be on the right? I think this would be especially beneficial on mobile, where the diamond currently ends up pretty far from the text. We could do this in a separate PR, let me know what you think.
> 
> CleanShot 2022-08-09 at 12 04 35@2x
> 
> CleanShot 2022-08-09 at 12 02 46@2x

I understand what you mean, but I am not sure if we should do this just for diamond icon, to me if we are doing this for diamond then shouldn't we do this for all icons? I think we should ask @ahmadmardeni1 for what he thinks to be honest. Because I don't know how would be the understanding arabic user in this all. @ahmadmardeni1 what do you recommend us doing please?

ahmadmardeni1:
Yes, all diamonds should be on the right side. But on staging everything is back to `ltr` for some reason:

Capture

Can you check so I can take a look? @alexandriaroberts

alexandriaroberts:
> Can you check so I can take a look? @alexandriaroberts
ohh, we need to redeploy the staging. Let me do that and I will message you when it's done @ahmadmardeni1. Thank you for checking this!

pages/_app.tsx
benface: The issue with the horizontal scroll is fixed by the `overflow: hidden` parent you added. :)

ahmadmardeni1:
I've just checked the new deployment. I think all icons are already on the right side like the search and GitHub icons except for the diamond and yes it should be on the right side. @alexandriaroberts @benface

ahmadmardeni1:
Just checked, great work! @alexandriaroberts

ahmadmardeni1 (approved):

benface (approved):
Ready to go! 🚀

PR #177, opened 2022-07-27
Add developer role to decentralized network

MichaelMacaulay:
Adding a page that's an overview of what developers do on the decentralized network.

PR #180, opened 2022-08-05
Created a Video component and updated the code with new component.

alexandriaroberts:
Added new video component to the docs, which takes `src` prop. Moved the styles onto the component in `sx` and updated the codebase with new Video component.

Screenshot 2022-08-05 at 16 09 00


components/Video.tsx
```diff
@@ -0,0 +1,16 @@
+import { HTMLAttributes } from 'react'
+
+export type VideoProps = { src: string } & HTMLAttributes<HTMLVideoElement>
```
hasparus, line 3: This seems to be an accident or an overeager GitHub Copilot suggestion.

image

We expose a type promising we'll accept the same props as `<video />` intrinsic element, so
if we pass an `onClick` we'll expect `event.currentTarget` to be `HTMLVideoElement` containing its usual props like `autoplay`.

But in the line 7, we're not spreading the props into a `<video />`, we have a `<figure />` there, so the `event.currentTarget` won't be `HTMLVideoElement`.

We can fix this easily by changing the type like this ⬇️

```suggestion
export type VideoProps = { src: string } & HTMLAttributes<HTMLElement>
```

It also seems non-obvious to me that `<Video />` does not render `<video />` inside.

As this component is only ever used with YouTube video embeds, what do you think about renaming it to **`<Embed />`**?

hasparus: As a side note, if we wanted to spread props to the intrinsic `<video />` element, we could use `React.ComponentPropsWithoutRef<'video'>` or `React.VideoHTMLAttributes<HTMLVideoElement>`.

image

hasparus (approved):
Nice work! I've left one small comment with regards to TypeScript and naming, but I love the cleanup.

components/Video.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+export const Video = ({ src, ...props }: VideoProps) => {
+  return (
+    <figure sx={{ paddingBottom: `${100 / (16 / 9)}%` }} {...props}>
+      <iframe
+        src={src}
+        frameBorder="0"
```
benface, line 10: Can you please check if this attribute changes anything (ideally in different browsers)? I have a feeling it's not necessary.

pages/ar/cookbook/migrating-a-subgraph.mdx
```diff
@@ -2,6 +2,8 @@
 title: ترحيل Subgraph موجود إلى شبكة The Graph
 ---
 
+import { Video } from '@/components'
```
benface, line 5: I just did a quick test and it looks like we can make the component global (avoid having to import it every time we need it) by importing it in `MDXLayout.tsx` and including it in the `mdxComponents` object (with `Video: Video,` or just `Video,`). Could you please do that instead? And while you're at it, maybe do it with `<Difficulty />` too? 😇

pages/ar/cookbook/migrating-a-subgraph.mdx
```diff
@@ -147,15 +149,7 @@ _بينما تتم هذه الآلية حاليا على الشبكة ، ينا
...
-<figure className="video-container">
-  <iframe
-    className="video-iframe"
-    src="https://www.youtube.com/embed/CzdQ3dFFrjo"
-    title="مشغل فيديو يوتيوب"
-    frameBorder="0"
-    allowFullScreen
-  ></iframe>
-</figure>
+<Video src="https://www.youtube.com/embed/CzdQ3dFFrjo" />
```
benface, line 152: So much better, right!?

components/Video.tsx
benface: How about `<VideoEmbed />`?

benface: `<Embed />` could mean something other than videos. Tweets can be embedded, for instance.

benface (changes requested):
Great work! Just a couple things: I agree with Piotr's comments and I added some of my own.

components/Video.tsx
alexandriaroberts: Yes, `<VideoEmbed />` would work. Also, for me having a video in the naming is self-explanatory that this is for video than anything else.

components/Video.tsx
alexandriaroberts: Yes, it is not necessary. Have removed it. :)

pages/ar/cookbook/migrating-a-subgraph.mdx
alexandriaroberts: yes, so much better! :)

components/Video.tsx
alexandriaroberts: Hey @hasparus , I have updated it to `<HTMLElement>` and thank you so much explaining it so well. I am still learning about types, and typescript really. So this was really helpful to understand why we need to use `<HTMLElement>` and not `<HTMLVideoElement>`

pages/ar/cookbook/migrating-a-subgraph.mdx
alexandriaroberts: Yes, that works very well and have done it for `<Difficulty />` too. Thank you! ☺️

components/Video.tsx
alexandriaroberts: What do you both think about `<EmbedVideo />` ? This could also work. Let me know which one we prefer so I will update it to that. Thank you.

benface: @alexandriaroberts – Personally, I prefer `VideoEmbed` since it reads better (like in a sentence, you would say "a video embed" or "an embedded video" but not "an embed video") + it starts with the most important thing about it IMO (at least from the content's perspective), which is that it renders a video.

hasparus: > which is that it renders a video.

well... it technically renders a website :P 

```
<Video src="https://twitframe.com/show?url=https%3A%2F%2Ftwitter.com%2Fjack%2Fstatus%2F20" />
```

This would work, right?

benface: > This would work, right?

@hasparus – Haha, yes, that would work. We could certainly add some validation for the `src` prop, but we wouldn't want to make it YouTube-only; for instance, we want to have [a Loom embed](https://github.com/graphprotocol/docs/pull/177/files#diff-6c9feb46d631ce72b595a06017b1048c88cb2566b4cfe7475f3aea3680a41f9fR52) on one page, and I could also see us wanting to embed a Vimeo or something else. It's not that component doesn't care about what it renders, but I don't see the point in _not_ having that flexibility; as long as it renders a 16:9 aspect ratio iframe (and we could make that customizable at some point, but it would definitely remain the default) I think we're good.

> I could also see a component like `<YouTubeEmbed video="dQw4w9WgXcQ" />`

That's a good idea! I like that it prevents a potential developer mistake, which is to copy the regular YouTube URL for a video instead of the embed one. If we follow a previously established convention, we could make it a more specific variant of `Video`, e.g. `<Video.YouTube video="dQw4w9WgXcQ" />`.

hasparus: Actually, maybe a `<VideoEmbed youtube="dQw4w9WgXcQ" />`?

benface: I like that! Then we can easily add "support" for other services.

alexandriaroberts: Okay, I understand what you mean.🙂 I have updated it will `VideoEmbed` and will push the code now.

hasparus (approved):
❤️

components/Video.tsx
benface: Thanks Lexie! I will add the `youtube` prop suggested by Piotr (as an alternative to `src` – if you set `src` it will override `youtube`). I think it's a great idea.

benface (approved):
🔥

PR #184, opened 2022-08-17
Removed padding for left sidebar elements

alexandriaroberts:
This doesn't have a task, removed padding on the sidebar, and the dropdown elements padding so it doesn't look so big and takes so much space.

Screenshot 2022-08-17 at 14 43 45
Screenshot 2022-08-17 at 14 43 56

### DIFF
--- a/components/NavTree.tsx
+++ b/components/NavTree.tsx
@@ -68,7 +68,7 @@ const NavTreeItem = ({
   const { sx: linkSx, ...linkOtherProps } = linkProps
   const { sx: diamondSx, ...diamondOtherProps } = diamondProps
   return (
-    <li sx={{ py: Spacing.M }} {...props}>
+    <li {...props}>
       <Link
         href={href}
         target={target}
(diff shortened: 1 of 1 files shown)


benface (approved):
This looks great, but if you'll allow me to be a little nitpicky, I noticed that the space above the "Subgraphs" heading hasn't changed so it now looks a bit too big, compared to the space between a divider and an actual nav item. 🤪

CleanShot 2022-08-17 at 19 40 48@2x

alexandriaroberts:
> This looks great, but if you'll allow me to be a little nitpicky, I noticed that the space above the "Subgraphs" heading hasn't changed so it now looks a bit too big, compared to the space between a divider and an actual nav item. 🤪
> 
Ohh I did that intentionally because I thought they were headings so maybe leave out more spacing for them and not nav items. But I will try removing it and see it looks nicer. Thank you. 🙂

benface (approved):

PR #193, opened 2022-09-02
Make external links open in a new tab by default

benface:
This PR changes links in `mdx` files to automatically have `target="_blank"` if the `href` doesn't start with `/` or `#` (meaning it's an external link, and we've been following the pattern of opening external links in new tabs on thegraph.com).


hasparus:
Technically, you could use two leading slashes in an external link, but I wrote `isExternal = href.startsWith('http')` so many times in my life that I lost the right to request changes on this PR.

https://stackoverflow.com/questions/9646407/two-forward-slashes-in-a-url-src-href-attribute

hasparus (approved):

benface:
> Technically, you could use two leading slashes in an external link

@hasparus – Super good catch, but it looks like even `next/link` doesn't support that. 😅 `<NextLink href="//thegraph.com" passHref>` renders `href="/thegraph.com"`.

Still, looking into this made me improve/simplify the code (removed 8 lines), so thank you.

PR #197, opened 2022-09-07
update description

0x2me:


azf20:
hey @tuoomz thanks - I think this is a question of taste, I have a preference for the existing wording

PR #204, opened 2022-09-19
Cost analysis tweak

MichaelMacaulay:
Clarifying info regarding engineering time


schmidsi (changes requested):
Can you try to preserve the table layout in the source file? I think that should be possible with Prettier: https://prettier.io

schmidsi (approved):

PR #209, opened 2022-09-27
Adding question

MichaelMacaulay:
Answering: "what is subgraph studio?"

### DIFF
--- a/pages/en/deploying/subgraph-studio-faqs.mdx
+++ b/pages/en/deploying/subgraph-studio-faqs.mdx
@@ -4,7 +4,7 @@ title: Subgraph Studio FAQs
 
 ### 1. What is Subgraph Studio?
 
-Subgraph Studio is a platform for creating, managing, and publishing subgraphs and API keys.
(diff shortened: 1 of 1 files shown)


pages/en/deploying/subgraph-studio-faqs.mdx
```diff
@@ -2,25 +2,29 @@
 title: Subgraph Studio FAQs
 ---
 
-### 1. How do I create an API Key?
+### 1. What is Subgraph Studio?
+
+Subgraph Studio is a platform for creating, managing, and publishing subgraphs and API keys.
```
benface, line 7: Is "platform" the right term? What about "web app" or "dapp"? Also, should we link to it (https://thegraph.com/studio/)? Seems odd not to.

MichaelMacaulay: I think dapp works best! Good call.

I'll add that link in too. Thanks!

benface (approved):

PR #210, opened 2022-09-28
Add Grafting Article

t-proctor:
Adding an article to the cookbook that describes grafting, as requested by @schmidsi

### DIFF
--- a/pages/en/cookbook/grafting.mdx
+++ b/pages/en/cookbook/grafting.mdx
@@ -1,0 +1,219 @@
+---
+title: Grafting on Existing Subgraphs
+---
+
+In this guide, you will learn how to build and deploy new subgraphs by grafting existing subgraphs.
+
+## What are Subgraphs?
+
+The Graph allows you to build custom open APIs called "subgraphs". Subgraphs are used to tell indexers (server operators) which data to index on a blockchain and save on their servers in order for you to be able to query it at any time using [GraphQL](https://graphql.org/).
+
+## What is Grafting?
+
+Grafting reuses the data from an existing subgraph and starts indexing it at a later block. This is useful during development to get past simple errors in the mappings quickly or to temporarily get an existing subgraph working again after it has failed
+
+The grafted subgraph can use a GraphQL schema that is not identical to the one of the base subgraph, but merely compatible with it. It has to be a valid subgraph schema in its own right, but may deviate from the base subgraph's schema in the following ways:
+
+- It adds or removes entity types
+- It removes attributes from entity types
+- It adds nullable attributes to entity types
+- It turns non-nullable attributes into nullable attributes
+- It adds values to enums
+- It adds or removes interfaces
+- It changes for which entity types an interface is implemented
+
+For more information, you can check:
+
+- [Grafting](https://thegraph.com/docs/en/developing/creating-a-subgraph/#grafting-onto-existing-subgraphs)
+
+> Note: Grafting requires that the Indexer has indexed the base subgraph. It is not recommended on The Graph Network at this time, and developers should not deploy subgraphs using that functionality to the network via the Studio.
+
+In this tutorial, we will be covering a basic usecase. We will replace an existing contract with an identical contract (with a new address). Then, graft the existing subgraph onto the new contract.
+
+## Building an Existing Subgraph
+
+To be able to build and deploy the existing subgraph used in this tutorial, you need to follow the instructions found here:
+
+1. `@graphprotocol/graph-cli`- This is a command-line tool for building and deploying subgraphs. [Click here](https://www.npmjs.com/package/@graphprotocol/graph-cli) to download using `npm`.
+
+Once these are installed, you can start building the subgraph with the instructions found here:
+
+- [Subgraph example repo](https://github.com/schmidsi/hackathon-starterkit#quick-start)
+
+> Note: The contract was taken from the following [Hackathon Starterkit](https://github.com/schmidsi/hackathon-starterkit).
+
+## Subgraph's Components
+
+There are three components of a subgraph:
+
+### 1. Manifest - `subgraph.yaml`
+
+Defines the data sources of interest, and how they should be processed. The Lock contract is our data source. We will be focusing on this file only in this tutorial.
+
+### 2. Schema - `schema.graphql`
+
+Here you define which data you want to be able to query after indexing your subgraph using GraphQL. This is actually similar to a model for an API, where the model defines the structure of a request body.
+
+The requirements for subgraphs are covered by the [existing documentation](/developing/creating-a-subgraph/#the-graphql-schema).
+
+### 3. AssemblyScript Mappings - `mapping.ts`
+
+This is the logic that determines how data should be retrieved and stored when someone interacts with the data sources you are listening to. The data gets translated and is stored based off the schema you have listed.
+
+During subgraph development there are two key commands:
+
+```
+$ graph codegen # generates types from the schema file identified in the manifest
+$ graph build # generates Web Assembly from the AssemblyScript files, and prepares all the subgraph files in a /build folder
+```
+
+## Subgraph Manifest Definition
+
+The subgraph manifest `subgraph.yaml` identifies the data sources for the subgraph, the triggers of interest, and the functions that should be run in response to those triggers. See below for an example subgraph manifest that you will use:
+
+```yaml
+specVersion: 0.0.4
+schema:
+  file: ./schema.graphql
+dataSources:
+  - kind: ethereum
+    name: Lock
+    network: goerli
+    source:
+      address: '0x4Ed995e775D3629b0566D2279f058729Ae6EA493'
+      abi: Lock
+      startBlock: 7674603
+    mapping:
+      kind: ethereum/events
+      apiVersion: 0.0.6
+      language: wasm/assemblyscript
+      entities:
+        - Withdrawal
+      abis:
+        - name: Lock
+          file: ./abis/Lock.json
+      eventHandlers:
+        - event: Withdrawal(uint256,uint256)
+          handler: handleWithdrawal
+      file: ./src/lock.ts
+```
+
+- The Locks' data source is the abi and contract address we will get when we compile and deploy the contract
+- The network should correspond to a indexed network being queried. Since we're running on Goerli testnet, the network is `goerli`
+- The `mapping` section defines the triggers of interest and the functions that should be run in response to those triggers. In this case, we are listening for the `Withdrawal` event and calling the `handleWithdrawal` function when it is emitted.
+
+## Grafting Manifest Definition
+
+Grafting requires adding two new items to the original subgraph manifest:
+
+```yaml
+---
+features:
+  - grafting # feature name
+graft:
+  base: Qm... # subgraph ID of base subgraph
+  block: 1502122 # block number
+```
+
+- `features:` is a list of all used [feature names](developing/creating-a-subgraph/#experimental-features).
+- `graft:` is a map of the `base` subgraph and the block to graft on to. The `block` is the block number to start indexing from. The Graph will copy the data of the base subgraph up to and including the given block and then continue indexing the new subgraph from that block on.
+
+The `base` and `block` values can be found by deploying two subgraphs: one for the base indexing and one with grafting
+
+## Deploying the Base Subgraph
+
+1. Go to the [Graph Studio UI](https://thegraph.com/studio/) and create a subgraph on Goerli testnet called `graft-example`
+2. Follow the directions in the `AUTH & DEPLOY` section on your subgraph page in the `graft-example` folder from the repo
+3. Once finished, verify the subgraph is indexing properly. If you run the following command in the graph playground
+
+```graphql
+{
+  withdrawals(first: 5) {
+    id
+    amount
+    when
+  }
+}
+```
+
+It should return the following:
+
+```
+{
+  "data": {
+    "withdrawals": [
+      {
+        "id": "0x13098b538a61837e9f29b32fb40527bbbe63c9120c250242b02b69bb42c287e5-5",
+        "amount": "0",
+        "when": "1664367528"
+      },
+      {
+        "id": "0x800c92fcc0edbd26f74e19ad058c62008a47c7789f2064023b987028343dd498-3",
+        "amount": "0",
+        "when": "1664367648"
+      }
+    ]
+  }
+}
+```
+
+Once you have verified the subgraph is indexing properly, you can deploy the subgraph with grafting.
+
+## Deploying the Grafting Subgraph
+
+1. Go to the [Graph Studio UI](https://thegraph.com/studio/) and create a subgraph on Goerli testnet called `graft-replacement`
+2. Add your `base` subgraph ID to the `graft` section in the `subgraph.yaml` file. This is the `Deployment ID` of your original `graph-example` subgraph. You can find this in the Graph Studio UI.
+3. Follow the directions in the `AUTH & DEPLOY` section on your subgraph page in the `graft-replacement` folder from the repo
+4. Once finished, verify the subgraph is indexing properly. If you run the following command in the graph playground
+
+```graphql
+{
+  withdrawals(first: 5) {
+    id
+    amount
+    when
+  }
+}
+```
+
+It should return the following:
+
+```
+{
+  "data": {
+    "withdrawals": [
+      {
+        "id": "0x13098b538a61837e9f29b32fb40527bbbe63c9120c250242b02b69bb42c287e5-5",
+        "amount": "0",
+        "when": "1664367528"
+      },
+      {
+        "id": "0x800c92fcc0edbd26f74e19ad058c62008a47c7789f2064023b987028343dd498-3",
+        "amount": "0",
+        "when": "1664367648"
+      },
+      {
+        "id": "0xb4010e4c76f86762beb997a13cf020231778eaf7c64fa3b7794971a5e6b343d3-22",
+        "amount": "0",
+        "when": "1664371512"
+      }
+    ]
+  }
+}
+```
+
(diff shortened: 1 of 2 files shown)


pages/en/cookbook/grafting.mdx
```diff
@@ -0,0 +1,219 @@
...
+
+There are three components of a subgraph:
+
+### 1. Manifest - `subgraph.yaml`
+
+Defines the data sources of interest, and how they should be processed. The Lock contract is our data source. We will be focusing on this file only in this tutorial.
+
+### 2. Schema - `schema.graphql`
+
+Here you define which data you want to be able to query after indexing your subgraph using GraphQL. This is actually similar to a model for an API, where the model defines the structure of a request body.
```
schmidsi, line 55: This is technically not correct. The schema simply defines the database schema how you want to store data. The resulting GraphQL schema is generated based on it.

pages/en/cookbook/grafting.mdx
```diff
@@ -0,0 +1,219 @@
...
+The requirements for subgraphs are covered by the [existing documentation](/developing/creating-a-subgraph/#the-graphql-schema).
+
+### 3. AssemblyScript Mappings - `mapping.ts`
+
+This is the logic that determines how data should be retrieved and stored when someone interacts with the data sources you are listening to. The data gets translated and is stored based off the schema you have listed.
+
+During subgraph development there are two key commands:
+
+```
+$ graph codegen # generates types from the schema file identified in the manifest
```
schmidsi, line 66: Also according to the loaded ABI

pages/en/cookbook/grafting.mdx
```diff
@@ -0,0 +1,219 @@
...
+$ graph build # generates Web Assembly from the AssemblyScript files, and prepares all the subgraph files in a /build folder
+```
+
+## Subgraph Manifest Definition
+
+The subgraph manifest `subgraph.yaml` identifies the data sources for the subgraph, the triggers of interest, and the functions that should be run in response to those triggers. See below for an example subgraph manifest that you will use:
```
schmidsi, line 72: I like "triggers of interest"

pages/en/cookbook/grafting.mdx
```diff
@@ -0,0 +1,219 @@
...
+{
+  withdrawals(first: 5) {
+    id
+    amount
+    when
+  }
+}
+```
+
+It should return the following:
```
schmidsi, line 139: It could be that this data has changed in the meantime. So I'd rather say: "It returns something like this"

pages/en/cookbook/grafting.mdx
```diff
@@ -0,0 +1,219 @@
...
+  }
+}
+```
+
+Once you have verified the subgraph is indexing properly, you can deploy the subgraph with grafting.
+
+## Deploying the Grafting Subgraph
+
+1. Go to the [Graph Studio UI](https://thegraph.com/studio/) and create a subgraph on Goerli testnet called `graft-replacement`
+2. Add your `base` subgraph ID to the `graft` section in the `subgraph.yaml` file. This is the `Deployment ID` of your original `graph-example` subgraph. You can find this in the Graph Studio UI.
```
schmidsi, line 165: We capitalise The Graph: The Graph Studio

pages/en/cookbook/grafting.mdx
```diff
@@ -0,0 +1,219 @@
...
+
+## Additional Resources
+
+If you want more experience with grafting, here's a few examples for popular contracts:
+
+- [Curve](https://github.com/messari/subgraphs/blob/master/subgraphs/curve-finance/protocols/curve-finance/templates/curve.template.yaml)
+- [ERC-721](https://github.com/messari/subgraphs/blob/master/subgraphs/erc721-metadata/subgraph.yaml)
+- [Uniswap](https://github.com/messari/subgraphs/blob/master/subgraphs/uniswap-v3/protocols/uniswap-v3/config/templates/uniswap.v3.template.yaml),
+
+To become even more of a Graph expert, consider learning about other ways to handle changes in underlying datasources. Alternatives like [Data Source Templates](developing/creating-a-subgraph/#data-source-templates) can achieve similar results
```
schmidsi, line 217: Yes!

pages/en/cookbook/grafting.mdx
schmidsi, line 165: Here I think we are missing the important points: How does the graft replacement subgraph.yaml differ from the original one?

pages/en/cookbook/grafting.mdx
```diff
@@ -0,0 +1,219 @@
...
+        "id": "0xb4010e4c76f86762beb997a13cf020231778eaf7c64fa3b7794971a5e6b343d3-22",
+        "amount": "0",
+        "when": "1664371512"
+      }
+    ]
+  }
+}
+```
+
+You can see that the `graft-replacement` subgraph is indexing from older `graph-example` data and newer data from the new contract address. The block number in `graft-replacement` specified in the `graft` section of the `subgraph.yaml` file has the last `Withdrawal` data from the old contract.
```
schmidsi, line 205: Also I think as a reader here I do not properly understand what's going on. You should clearly say that the first contract emitted events 1 & 2 and the new contract emitted 3. Maybe link to Etherscan so people can follow and verify.

schmidsi (changes requested):

pages/en/cookbook/grafting.mdx
```diff
@@ -0,0 +1,219 @@
+---
+title: Grafting on Existing Subgraphs
+---
+
+In this guide, you will learn how to build and deploy new subgraphs by grafting existing subgraphs.
+
+## What are Subgraphs?
```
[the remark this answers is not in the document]
t-proctor: Removed it

pages/en/cookbook/grafting.mdx
```diff
@@ -0,0 +1,219 @@
...
+- It turns non-nullable attributes into nullable attributes
+- It adds values to enums
+- It adds or removes interfaces
+- It changes for which entity types an interface is implemented
+
+For more information, you can check:
+
+- [Grafting](https://thegraph.com/docs/en/developing/creating-a-subgraph/#grafting-onto-existing-subgraphs)
+
+> Note: Grafting requires that the Indexer has indexed the base subgraph. It is not recommended on The Graph Network at this time, and developers should not deploy subgraphs using that functionality to the network via the Studio.
```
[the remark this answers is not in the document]
t-proctor: Removed it

pages/en/cookbook/grafting.mdx
t-proctor: Mentioned this

pages/en/cookbook/grafting.mdx
t-proctor: Fixed this wording

schmidsi (approved):
Looks good to me

pages/en/cookbook/grafting.mdx
```diff
@@ -0,0 +1,215 @@
+---
+title: Grafting on Existing Subgraphs
```
schmidsi, line 2: Maybe we can change the title to: "Replacing a contract address and keep history with Grafting"

azf20: I think my reservation about that title suggestion is that I don't think Grafting is the best way to do that (vs. adding a new data source with the new address). Grafting is a means of quickly fixing (vs. a preferred upgrade pattern)

pages/en/cookbook/grafting.mdx
```diff
@@ -0,0 +1,215 @@
...
+
+- [Grafting](https://thegraph.com/docs/en/developing/creating-a-subgraph/#grafting-onto-existing-subgraphs)
+
+In this tutorial, we will be covering a basic usecase. We will replace an existing contract with an identical contract (with a new address). Then, graft the existing subgraph onto the new contract.
+
+## Building an Existing Subgraph
+
+To be able to build and deploy the existing subgraph used in this tutorial, you need to follow the instructions found here:
+
+1. `@graphprotocol/graph-cli`- This is a command-line tool for building and deploying subgraphs. [Click here](https://www.npmjs.com/package/@graphprotocol/graph-cli) to download using `npm`.
```
azf20, line 31: I don't know if we want to promote global usage of `graph-cli` as a pattern?

pages/en/cookbook/grafting.mdx
```diff
@@ -0,0 +1,215 @@
...
+
+Once these are installed, you can start building the subgraph with the instructions found here:
+
+- [Subgraph example repo](https://github.com/t-proctor/grafting-tutorial)
+
+> Note: The contract was taken from the following [Hackathon Starterkit](https://github.com/schmidsi/hackathon-starterkit).
+
+## Subgraph's Components
+
+There are three components of a subgraph:
```
azf20, line 41: I don't know if we want to restate this information here, I think we should be able to redirect folks to the existing quick start

pages/en/cookbook/grafting.mdx
```diff
@@ -0,0 +1,215 @@
...
+        "id": "0x800c92fcc0edbd26f74e19ad058c62008a47c7789f2064023b987028343dd498-3",
+        "amount": "0",
+        "when": "1664367648"
+      }
+    ]
+  }
+}
+```
+
+Once you have verified the subgraph is indexing properly, you can deploy the subgraph with grafting.
```
azf20, line 154: I might reword this to "you can quickly update the subgraph with grafting"

pages/en/cookbook/grafting.mdx
t-proctor: ok, just changed it

azf20 (changes requested):
Thanks so much @t-proctor just some small changes here

pages/en/cookbook/grafting.mdx
```diff
@@ -0,0 +1,215 @@
...
+- It removes attributes from entity types
+- It adds nullable attributes to entity types
+- It turns non-nullable attributes into nullable attributes
+- It adds values to enums
+- It adds or removes interfaces
+- It changes for which entity types an interface is implemented
+
+For more information, you can check:
+
+- [Grafting](https://thegraph.com/docs/en/developing/creating-a-subgraph/#grafting-onto-existing-subgraphs)
```
marceljay, line 23: The link generator in the docs inserts a slash, which makes the the anchors not work.

```suggestion
- [Grafting](https://thegraph.com/docs/en/developing/creating-a-subgraph#grafting-onto-existing-subgraphs)
```

pages/en/cookbook/grafting.mdx
```diff
@@ -0,0 +1,215 @@
+---
+title: Replace a Contract and Keep its History With Grafting
+---
+
+In this guide, you will learn how to build and deploy new subgraphs by grafting existing subgraphs.
+
+## What is Grafting?
+
+Grafting reuses the data from an existing subgraph and starts indexing it at a later block. This is useful during development to get past simple errors in the mappings quickly or to temporarily get an existing subgraph working again after it has failed. Also, it can be used when adding a feature to a subgraph that takes long to index from scratch.
```
marceljay, line 9: The docs state:
> When a subgraph whose manifest contains a graft block is deployed, Graph Node will copy the data of the base subgraph up to and including the given block and then continue indexing the new subgraph from that block on. 

Does the base subgraph, onto which the one you currently running is grafted, have an end-block number? Intuitively I would say, it would have a start and end block.

pages/en/cookbook/grafting.mdx
```diff
@@ -0,0 +1,215 @@
...
+- It turns non-nullable attributes into nullable attributes
+- It adds values to enums
+- It adds or removes interfaces
+- It changes for which entity types an interface is implemented
+
+For more information, you can check:
+
+- [Grafting](https://thegraph.com/docs/en/developing/creating-a-subgraph/#grafting-onto-existing-subgraphs)
+
+In this tutorial, we will be covering a basic usecase. We will replace an existing contract with an identical contract (with a new address). Then, graft the existing subgraph onto the new contract.
```
marceljay, line 25:
```suggestion
In this tutorial, we will be covering a basic usecase. We will replace an existing contract with an identical contract (with a new address). Then, graft the existing subgraph onto the "base" subgraph that tracks the new contract.
```

Also not sure what you mean by "replace existing contract". Do you mean inside the manifest or on-chain as in a contract migration from address A to B.

pages/en/cookbook/grafting.mdx
```diff
@@ -0,0 +1,215 @@
...
+```
+
+Once you have verified the subgraph is indexing properly, you can deploy the subgraph with grafting.
+
+## Deploying the Grafting Subgraph
+
+The graft replacement subgraph.yaml will have a new contract address
+
+1. Go to [The Graph Studio UI](https://thegraph.com/studio/) and create a subgraph on Goerli testnet called `graft-replacement`
+2. Create a new manifest. The `subgraph.yaml` for `graph-replacement` contains a different contract address and new information about how it should graft. These are the `block` of the [last event emitted](https://goerli.etherscan.io/tx/0x800c92fcc0edbd26f74e19ad058c62008a47c7789f2064023b987028343dd498) you care about by the old contract and the `base` of the old subgraph. The `base` subgraph ID is the `Deployment ID` of your original `graph-example` subgraph. You can find this in The Graph Studio UI.
```
marceljay, line 161: I'm also not sure how the events are handled. In our case, we have the same contract address (UUPS Proxy Pattern) but a different Event Signature. Would that require us to add a separate Event Handler to the base manifest?

pages/en/cookbook/grafting.mdx
t-proctor: Just clarifying, are you suggesting to remove the instructions to install `graph-cli` entirely?

pages/en/cookbook/grafting.mdx
t-proctor: Could you clarify what you mean? Wouldn't the same event handler be fine since no code has changed, even though the addresses ([1](https://goerli.etherscan.io/address/0x4ed995e775d3629b0566d2279f058729ae6ea493),[2](https://goerli.etherscan.io/address/0xdb50720d920623a5bce4ffb65a06185dbf4d4c83)) are different?

pages/en/cookbook/grafting.mdx
t-proctor: Yes, are you suggesting adding this info to the beginning?

pages/en/cookbook/grafting.mdx
t-proctor: Thanks! Have added more to explain this

t-proctor:
@azf20 @marceljay thank you for the feedback, have added fixes/comments based on it.

pages/en/cookbook/grafting.mdx
marceljay: The contract code has changed, the new implementation (after upgrade) has different bytecode and different events.  IIUC a handler and an event have a 1:1 relationship. 
So the first version of the contract, the event had the following signature:
```
        - event: Retired(address,uint256)
```

But now it's:
```
        - event: Retired(address,uint256,uint256)
```
Would that not mean we need a different handler?

pages/en/cookbook/grafting.mdx
marceljay: In a regular subgraph, you have startblock, and from that onwards, the contract is being indexed. A base subgraph though, should not be indexed anymore after a certain block number, right?
The docs state the following:
image
Which makes it seem that the base is going to be indexed indefinitely too.

pages/en/cookbook/grafting.mdx
t-proctor: The event signature `event Withdrawal(uint256 amount, uint256 when);` is the same for both, so do not think it would change

pages/en/cookbook/grafting.mdx
t-proctor: Don't think [endblock feature](https://github.com/graphprotocol/graph-node/issues/3504) exists yet, but thanks for the suggestions!

pages/en/cookbook/grafting.mdx
marceljay: I meant in our case, having two different signatures with the same name (due to a contract upgrade), is the main reason we are looking into Grafting. It would be great if you could cover that use case in the tutorial 🙏

azf20: @marceljay in your case I don't think you need Grafting, you can just add a new event handler with the new signature

azf20 (approved):

azf20:
Thanks @t-proctor!

pages/en/cookbook/grafting.mdx
marceljay: It worked via Grafting. The old base subgraph uses a different handler implementation but maps to the same entity. This seems to work fine. I don't think it would work without it.

marceljay: @azf20 @t-proctor can you confirm the above? Meaning, that the code generate by `graph-cli` would overwrite the second event (with the same name), meaning it's not polymorphic?

azf20: hey @marceljay it is hard to confirm what is happening with only the snippets, are you able to share the full codebase?
I think the cleanest approach here with the upgrade is probably to add a new dataSource with the new ABI (pointing at the same address) and the new handler, rather than merging into the existing dataSource

aspiers: This approach (two dataSources) wouldn't require grafting, right?

azf20: yes no grafting required (this thread has gotten slightly off the original topic!)

marceljay: Thanks a lot, we can evaluate this internally. This would mean we need to have the legacy ABI in our build pipeline that is sharing artifacts within the mono-repo. I'd say there are pro's and con's.

PR #211, opened 2022-09-28
Small feature usage suggestion in quick-start

t-proctor:
Suggest adding a mention of a great feature, `--index-events` in the quick-start article


schmidsi (changes requested):
I really like this note. That said, I think we should clearly say what it does: It bootstraps the subgraph with entities in the schema and simple mappings for each emitted event.

t-proctor:
> I really like this note. That said, I think we should clearly say what it does: It bootstraps the subgraph with entities in the schema and simple mappings for each emitted event.

Added this, thanks!

azf20 (approved):

schmidsi (approved):

PR #205, opened 2022-09-20
Small updates to indexing

MichaelMacaulay:
Keeping wording consistent

PR #214, opened 2022-10-03
Adding Indexer options

MichaelMacaulay:
Adding a disclaimer that Indexers can use AWS, Alibaba, or MS Azure in place of Google Cloud


benface (commented):
Looks like this PR changes some of the same lines as #205, but differently (I think #205 is correct and this one just has these changes because of Prettier). Probably best to wait for #205 to be merged and then "use theirs" when resolving merge conflicts here.

PR #220, opened 2022-10-14
Removing recommendation against grafting

t-proctor:
The current documentation has a recommendation to not use grafting. Should this be removed as it has been functional for enough time now?


azf20 (approved):
This seems reasonable to me. Thanks.

PR #222, opened 2022-10-17
Updating Indexer Options

MichaelMacaulay:
Mentioning that Indexers can use alternatives to Google Cloud to run their indexing node.


MichaelMacaulay:
@azf20 In the previous PR, you asked if Alibaba was a common choice. Consensus was that since it's a "tier 1" service, we should include it. Especially because it is really popular in Asia.

azf20 (approved):

PR #230, opened 2022-11-10
Azf20/firehose and substreams

azf20:
- remove "The" from Firehose

### DIFF
--- a/navigation/navigation.ts
+++ b/navigation/navigation.ts
@@ -186,4 +186,13 @@ export const navigation = (locale: AppLocale): NavItemDefinition[] => [
       },
     ],
   },
+  {
+    divider: true,
+  },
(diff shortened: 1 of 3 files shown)


benface:
@azf20 – Looks like you used the same branch that was merged (then auto-deleted) in #228, but we squash commits so even though you had a single commit, it was rewritten by GitHub as a different commit (with the title of the PR as its message), which means that it's now seen as a new commit and it conflicts with the current state of `main`. What I like to do when this happens is cherry-pick the actual new commits in another branch and force-push to the remote branch (or create another PR for that new branch). Or, you can resolve the conflict manually.

azf20:
@benface good point, I rebased

benface (approved):

PR #231, opened 2022-11-10
modify Access Smart Contract State

itsjerryokolo:
The modifications I made to `Access Smart Contract State` were pulled from this [section](https://thegraph.com/docs/en/developing/assemblyscript-api/#creating-entities)

In its current form, there's a conflict between the entity name and event name.


pages/en/developing/assemblyscript-api.mdx
```diff
@@ -444,19 +444,20 @@ The code generated by `graph codegen` also includes classes for the smart contra
 A common pattern is to access the contract from which an event originates. This is achieved with the following code:
 
 ```typescript
-// Import the generated contract class
-import { ERC20Contract } from '../generated/ERC20Contract/ERC20Contract'
+// Import the generated contract class and generated Transfer event class
+import { ERC20Contract, Transfer as TransferEvent } from '../generated/ERC20Contract/ERC20Contract'
 // Import the generated entity class
 import { Transfer } from '../generated/schema'
```
azf20, line 450: I think this line can actually be deleted for the purpose of this snippet?

azf20 (approved):
Looks good! I think actually in this case we don't need to import the Transfer from the schema, but maybe that is a good pointer to have in there

pages/en/developing/assemblyscript-api.mdx
itsjerryokolo: Line 450?

PR #233, opened 2022-11-22
Add common AssemblyScript issues in the scope of subgraphs

melotik:
@graphprotocol/docs-reviewers

### DIFF
--- a/pages/en/developing/assemblyscript-api.mdx
+++ b/pages/en/developing/assemblyscript-api.mdx
@@ -761,3 +761,14 @@ The base `Entity` class and the child `DataSourceContext` class have helpers to 
 - `getBytes(key: string): Bytes`
 - `getBoolean(key: string): boolean`
 - `getBigDecimal(key: string): BigDecimal`
+
+### Common AssemblyScript Issues
+
+There are certain [AssemblyScript](https://github.com/AssemblyScript/assemblyscript) issues that are common to run into during subgraph development. They range in debug difficulty, however, being aware of them may help. The following is a non-exhaustive list of these issues:
+
+- Be careful with handler computation limits, especially when using the `.pow()` function with large numbers.
+  - As of [`graph-node v0.26.0`](https://github.com/graphprotocol/graph-node/releases/tag/v0.26.0) gas metering was introduced.
+- Required fields are handled differently depending on the `graph-cli` version used.
+  - In older `graph-cli` releases default values are automatically set, but this was changed to require non-nullable fields to be set in [`graph-cli vTODO?`]().
+- `Private` class variables are not enforced in [AssembyScript](https://www.assemblyscript.org/status.html#language-features). There is no way to protect class variables from being directly changed from the class object.
+- Scope is not inherited into [closure functions](https://www.assemblyscript.org/status.html#on-closures), i.e. variables declared outside of closure functions cannot be used. Explanation in [Developer Highlights #3](https://www.youtube.com/watch?v=1-8AW-lVfrA&t=3243s).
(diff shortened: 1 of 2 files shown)


pages/en/developing/creating-a-subgraph.mdx
```diff
@@ -303,6 +303,8 @@ Reverse lookups can be defined on an entity through the `@derivedFrom` field. Th
 
 For one-to-many relationships, the relationship should always be stored on the 'one' side, and the 'many' side should always be derived. Storing the relationship this way, rather than storing an array of entities on the 'many' side, will result in dramatically better performance for both indexing and querying the subgraph. In general, storing arrays of entities should be avoided as much as is practical.
 
+> Note: You cannot use `@derivedFrom` fields in subgraph mappings. It will compile, but throw an `unexpected null wasm` error during runtime. See: [#219](https://github.com/graphprotocol/graph-ts/issues/219)
```
azf20, line 306: we are actually working on changing this https://github.com/graphprotocol/graph-node/pull/4119

pages/en/developing/assemblyscript-api.mdx
```diff
@@ -761,3 +761,14 @@ The base `Entity` class and the child `DataSourceContext` class have helpers to
 - `getBytes(key: string): Bytes`
 - `getBoolean(key: string): boolean`
 - `getBigDecimal(key: string): BigDecimal`
+
+### Common AssemblyScript Issues
+
+There are certain [AssemblyScript](https://github.com/AssemblyScript/assemblyscript) issues that are common to run into during subgraph development. They range in debug difficulty, however, being aware of them may help. The following is a non-exhaustive list of these issues:
+
+- Be careful with handler computation limits, especially when using the `.pow()` function with large numbers.
+  - As of [`graph-node v0.26.0`](https://github.com/graphprotocol/graph-node/releases/tag/v0.26.0) gas metering was introduced.
```
azf20, line 770: I don't think the goal here was to penalise reasonable actions in mappings (cc @leoyvens), can you give more detail on when you encountered an issue with `.pow()` (the gas setting might not be appropriate)

pages/en/developing/assemblyscript-api.mdx
```diff
@@ -761,3 +761,14 @@ The base `Entity` class and the child `DataSourceContext` class have helpers to
...
+- Required fields are handled differently depending on the `graph-cli` version used.
+  - In older `graph-cli` releases default values are automatically set, but this was changed to require non-nullable fields to be set in [`graph-cli vTODO?`]().
```
azf20, line 772: This was actually an inadvertent codegen issue which didn't exist in oldest versions, but was introduced temporarily. The reason this was reverted back was to support the `upsert` behaviour in Graph Node when you create an entity that already exists. I think in general we should encourage usage of the most recent versions, so I am not sure if we want to phrase this in this way?

azf20: I think we probably want to be more definitive to explain the current behaviour, with a caveat that in some prior versions of graph-cli this was handled differently

pages/en/developing/assemblyscript-api.mdx
leoyvens: There once was a gas issue with `pow` but that was fixed. The issue with abusing `pow` on `BigInt` is that it can cause the number to quickly grow in size.

dmelotik: removed

pages/en/developing/assemblyscript-api.mdx
dmelotik: removed from here, should this be added somewhere else? But to be clear, currently null fields that are not set have defaults and will not fail to index?

pages/en/developing/creating-a-subgraph.mdx
dmelotik: removed

pages/en/developing/assemblyscript-api.mdx
azf20: sorry I was unclear - null fields don't have defaults, they need to be set (though I think we should ideally catch this during `build` rather than runtime)

dmelotik: Do you think these additions fit here? To me it feels somewhat out of place. It's really just assembly script quirks that we have to be cognizant of during subgraph development.

Web3Slimchance (approved):
Looks good to me

PR #240, opened 2022-11-24
Update operating-graph-node.mdx

Web3Slimchance:
- Added another case where `trace_filter` is required 
- Added codeblock format to `toml` snippets
- Changed the order of log levels to be more intuitive (ordered from least to most logs)
- Removed link to internal Grafana.  
- Fixed a couple Typos


azf20 (approved):
Thanks for the fixes!

PR #241, opened 2022-12-01
Update dependencies + quick fixes

benface:
Includes this fix: https://github.com/edgeandnode/components/pull/542


cmwhited (approved):
How does lodash not ship with types in dang near 2023. smh

PR #245, opened 2022-12-08
Fixing typo

MichaelMacaulay:


pages/en/querying/graphql-api.mdx
```diff
@@ -205,7 +205,7 @@ _change_block(number_gte: Int)
 
 ### Time-travel queries
 
-You can query the state of your entities not just for the latest block, which is the by default, but also for an arbitrary block in the past. The block at which a query should happen can be specified either by its block number or its block hash by including a `block` argument in the toplevel fields of queries.
+You can query the state of your entities not just for the latest block, which is by default, but also for an arbitrary block in the past. The block at which a query should happen can be specified either by its block number or its block hash by including a `block` argument in the toplevel fields of queries.
```
benface, line 208:
```suggestion
You can query the state of your entities not just for the latest block, which is the default, but also for an arbitrary block in the past. The block at which a query should happen can be specified either by its block number or its block hash by including a `block` argument in the toplevel fields of queries.
```
🤔

MichaelMacaulay: Makes more sense your way - thanks!

benface (approved):

PR #258, opened 2023-01-03
fix studio billing urls to prod from staging

nasmr:
Fix for billing page, where Studio hyperlinks were incorrectly pointing to Staging env (https://staging.thegraph.com/studio/billing) instead of Production billing page (https://thegraph.com/studio/billing)

@graphprotocol/docs-reviewers


benface (approved):
😅

PR #274, opened 2023-01-30
some small fixes that I found

dimaMachina:


.gitignore
```diff
@@ -36,3 +36,5 @@ yarn-error.log*
 
 # typescript
 *.tsbuildinfo
+
+.idea/
```
B2o5T, line 40: to ignore my default webstorm settings folder

components/Paragraph.tsx
```diff
@@ -10,7 +10,7 @@ export const Paragraph = ({ children, ...props }: ParagraphProps) => {
       sx={{
         mt: Spacing['16px'],
         mb: Spacing['24px'],
-        '&:first-child': { mt: 0 },
+        '&:first-of-type': { mt: 0 },
```
B2o5T, line 13: image

pages/_app.tsx
```diff
@@ -30,7 +30,7 @@ const DEFAULT_SEO_PROPS: DefaultSeoProps = {
   title: 'The Graph Docs',
   description: 'Browse the latest developer documentation including API reference, articles, and sample code',
   openGraph: {
-    site_name: 'The Graph Docs',
+    siteName: 'The Graph Docs',
```
B2o5T, line 33: image

saihaj (approved):

components/Paragraph.tsx
benface: Hmm, these are not exactly equivalent, but I just tested and it doesn't break anything... nor does removing this entire rule... or the following one (the `&:last-child: { mb: 0 }` one). So, I propose doing that. 😄

components/Paragraph.tsx
```diff
@@ -10,7 +10,7 @@ export const Paragraph = ({ children, ...props }: ParagraphProps) => {
...
         '&:last-child': { mb: 0 },
```
benface, line 14:
```suggestion
```

benface (approved):
Thank you!

PR #275, opened 2023-01-30
add remote mdx

dimaMachina:
fetched remotly

image

### DIFF
--- a/lib/remarkMdxLayout.mjs
+++ b/lib/remarkMdxLayout.mjs
@@ -30,7 +30,7 @@ export const remarkMdxLayout = () => {
   
           const outline = ${Serializer.run(outline)}
   
-          export const getStaticProps = async () => {
+          export const getStaticProps = async (params) => {
             const locale = '${locale}'
             const navItems = await getNavItems(locale)
 
@@ -38,6 +38,7 @@ export const remarkMdxLayout = () => {
               props: {
                 locale,
                 navItems,
+                ...(typeof mdxGetStaticProps !== 'undefined' && (await mdxGetStaticProps(params)).props)
               },
             }
           }
--- a/components/SSG.ts
+++ b/components/SSG.ts
@@ -1,0 +1,5 @@
+import { createContext, useContext } from 'react'
+
+export const SSGContext = createContext<any>(false)
+
+export const useSSG = (key = 'ssg') => useContext(SSGContext)?.[key]
--- a/package.json
+++ b/package.json
@@ -32,6 +32,7 @@
     "@reach/auto-id": "^0.18.0",
     "lodash": "^4.17.21",
     "next": "^12.3.4",
+    "next-mdx-remote": "^4.3.0",
     "next-seo": "^5.15.0",
     "prism-react-renderer": "^1.3.5",
     "react": "^17.0.2",
--- a/pages/_app.tsx
+++ b/pages/_app.tsx
@@ -19,6 +19,7 @@ import {
   ThemeProvider,
 } from '@edgeandnode/components'
 
+import { SSGContext } from '@/components/SSG'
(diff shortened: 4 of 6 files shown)


lib/remarkMdxLayout.mjs
```diff
@@ -30,14 +30,15 @@ export const remarkMdxLayout = () => {
...
-          export const getStaticProps = async () => {
+          export const getStaticProps = async (params) => {
             const locale = '${locale}'
             const navItems = await getNavItems(locale)
 
             return {
               props: {
                 locale,
                 navItems,
+                ...(typeof mdxGetStaticProps !== 'undefined' && (await mdxGetStaticProps(params)).props)
```
B2o5T, line 41: since we can't have 2 `getStaticProps` on the page I made some workaround 
image

dotansimha (approved):

benface (commented):
This is a nice and clever implementation, but won't it lead to a lot of duplicated boilerplate if we have multiple pages that need to render remote content? Technically, the only thing in `remote.mdx` that we'd want to change is the `https://raw.githubusercontent.com/shuding/nextra/main/README.md` URL, isn't it? Now, maybe this isn't a good idea, but I'm throwing it out there: what if we added a `fetch` frontmatter property that, when set, automatically generated all that boilerplate (likely in `remarkMdxLayout.mjs`)?

For example, `remote.mdx` could be:
```mdx
---
title: Remote MDX Example
fetch: https://raw.githubusercontent.com/shuding/nextra/main/README.md
---
```

And if we push that idea further, we could support multiple named fetches and allow rendering them where we want:
```mdx
---
title: Remote MDX Example
fetch:
  remoteContent1: https://raw.githubusercontent.com/shuding/nextra/main/README.md
  remoteContent2: https://raw.githubusercontent.com/shuding/nextra/main/CODE_OF_CONDUCT.md
---

## Lorem ipsum

<Remote content={remoteContent1} />

## Lorem ipsum

<Remote content={remoteContent2} />
```

lib/remarkMdxLayout.mjs
```diff
@@ -30,14 +30,15 @@ export const remarkMdxLayout = () => {
   
           const outline = ${Serializer.run(outline)}
   
-          export const getStaticProps = async () => {
+          export const getStaticProps = async (params) => {
```
benface, line 33: Should probably be `context` instead of `params`. Also, this is just in case `mdxGetStaticProps` might need the context (which doesn't seem to be the case in your example), right?

B2o5T:
> This is a nice and clever implementation, but won't it lead to a lot of duplicated boilerplate if we have multiple pages that need to render remote content?

@benface you are right. Migration to nextra will fix it

Example https://nextra-theme-docs-ojscqyf0a.vercel.sh/remote/graphql-eslint

All remote files for `graphql-eslint` and `graphql-yoga` were fetched and built with few lines of code and a valid sidebar was generated without forgetting of including some files :)
https://github.com/shuding/nextra/tree/main/examples/swr-site/pages/remote/graphql-eslint
https://github.com/shuding/nextra/tree/main/examples/swr-site/pages/remote/graphql-yoga

lib/remarkMdxLayout.mjs
B2o5T: yeah but I added it in case some another mdxGetStaticProps will need it in the future ;)

benface:
That's really cool @B2o5T. Do you think it's worth adding the remote content that we want to include in the docs with this stopgap solution, or do you think we should move to Nextra ASAP and this was more a proof of concept?

B2o5T:
@benface I guess remote mdx is better to add after Nextra migration to avoid DRY and keep the code clean and scalable.

B2o5T:
closing in favor https://github.com/graphprotocol/docs/pull/301

PR #289, opened 2023-02-23
Add Staking.abi URL

mgarciap:
It was missing


MichaelMacaulay (approved):
Great catch! Thanks so much!

PR #291, opened 2023-02-23
Update creating-a-subgraph.mdx

kev1n:
Updated the list of unsupported Call Handler chains.

Was wondering why my subgraphs were failing with a NodeCapabilities error. Hopefully, this helps other devs find the solution faster.

### DIFF
--- a/pages/en/developing/creating-a-subgraph.mdx
+++ b/pages/en/developing/creating-a-subgraph.mdx
@@ -712,7 +712,7 @@ While events provide an effective way to collect relevant changes to the state o
 
 Call handlers will only trigger in one of two cases: when the function specified is called by an account other than the contract itself or when it is marked as external in Solidity and called as part of another function in the same contract.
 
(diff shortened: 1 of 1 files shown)


pages/en/developing/creating-a-subgraph.mdx
```diff
@@ -712,7 +712,7 @@ While events provide an effective way to collect relevant changes to the state o
 
 Call handlers will only trigger in one of two cases: when the function specified is called by an account other than the contract itself or when it is marked as external in Solidity and called as part of another function in the same contract.
 
-> **Note:** Call handlers currently depend on the Parity tracing API. Certain networks, such as BNB chain and Arbitrum, does not support this API. If a subgraph indexing one of these networks contain one or more call handlers, it will not start syncing. Subgraph developers should instead use event handlers. These are far more performant than call handlers, and are supported on every evm network.
+> **Note:** Call handlers currently depend on the Parity tracing API. Certain networks, such as BNB chain and Arbitrum, Polygon, and Optimism do not support this API. If a subgraph indexing one of these networks contain one or more call handlers, it will not start syncing with a NodeCapabilities error. Subgraph developers should instead use event handlers. These are far more performant than call handlers, and are supported on every evm network.
```
benface, line 715:
```suggestion
> **Note:** Call handlers currently depend on the Parity tracing API. Certain networks, such as BNB chain, Arbitrum, Polygon, and Optimism do not support this API. If a subgraph indexing one of these networks contain one or more call handlers, it will not start syncing with a NodeCapabilities error. Subgraph developers should instead use event handlers. These are far more performant than call handlers, and are supported on every evm network.
```

PR #296, opened 2023-02-27
chore: move to monorepo

saihaj:
Move to a monorepo structure. All the docs live in `website/*` so in future we can add `packages/*` so we can maintain shared configs, components and themes for Nextra.


saihaj:
we can ignore this comment since right now on `main` there is nothing like `/website/route-lockfile.txt`

.gitignore
```diff
@@ -1,5 +1,5 @@
 # dependencies
-/node_modules
+**/node_modules
```
B2o5T, line 2:
```suggestion
node_modules/
```

website/package.json
```diff
@@ -0,0 +1,60 @@
...
+    "remark-frontmatter": "^4.0.1",
+    "remark-mdx-frontmatter": "^2.1.1",
+    "serialize-as-code": "^2.0.2",
+    "typescript": "4.9.5",
+    "unist-util-visit": "^4.1.2",
+    "url": "^0.11.0"
+  },
+  "lint-staged": {
+    "**/*.{js,jsx,ts,tsx,mjs}": "eslint --fix",
+    "**/*.{js,jsx,ts,tsx,mjs,mdx,json}": "prettier --write"
```
B2o5T, line 58:
```suggestion
    "**/*.{js,jsx,ts,tsx,mjs,cjs}": "eslint --fix",
    "**/*.{js,jsx,ts,tsx,mjs,cjs,md,mdx,json}": "prettier --write"
```

dimaMachina (approved):

dotansimha (approved):

website/package.json
```diff
@@ -0,0 +1,60 @@
+{
+  "name": "@graphprotocol/docs",
+  "version": "1.0.0",
+  "packageManager": "pnpm@7.28.0",
+  "private": true,
+  "scripts": {
+    "dev": "next dev",
+    "build": "rm -rf .next && next build",
+    "export": "rm -rf out && next export -o ../out/docs",
```
benface, line 9:
```suggestion
    "export": "rm -rf ../out && next export -o ../out/docs",
```
Though, I wonder...
1. Should `out` be under `website`? (That would require changing some stuff in `Dockerfile`, but...)
2. Should `Dockerfile` be under `website`?

pnpm-workspace.yaml
```diff
@@ -0,0 +1,3 @@
+packages:
+  - packages/*
+  - website
```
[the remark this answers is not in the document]
B2o5T: we can put it in `packages` 👍

website/package.json
```diff
@@ -0,0 +1,60 @@
+{
+  "name": "@graphprotocol/docs",
+  "version": "1.0.0",
+  "packageManager": "pnpm@7.28.0",
```
B2o5T, line 4: I guess we don't need it here since we already specified it in root package.json

package.json
```diff
@@ -21,53 +19,15 @@
...
-    "@svgr/webpack": "^6.5.1",
-    "@types/color": "^3.0.3",
-    "@types/lodash": "^4.14.191",
-    "@types/node": "^16.18.12",
-    "@types/react": "^17.0.53",
-    "@types/react-dom": "^17.0.18",
-    "acorn": "^8.8.2",
-    "acorn-jsx": "^5.3.2",
     "eslint": "^8.34.0",
     "fast-xml-parser": "^4.1.2",
```
B2o5T, line 25: it's used only in `website` can you move in the corresponding `package.json`?

B2o5T:
can you also undo moving `.eslintrc.js` to a `website` and keep it in root?

website/package.json
B2o5T: I think also `Dockerfile` should be in `website`!

package.json
```diff
@@ -21,53 +19,15 @@
...
     "husky": "^8.0.3",
     "lint-staged": "^13.1.1",
     "path": "^0.12.7",
```
B2o5T, line 28: @benface any thoughts why we need this package?

pnpm-workspace.yaml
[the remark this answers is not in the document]
saihaj: it is simpler to keep on root. This makes it so that we can have docker context bootstrap from the root of mono-repo. We can build the full monorepo from root and then only have it copy files that matter in final image that we publish.

website/package.json
saihaj: it is simpler to keep on root. This makes it so that we can have docker context bootstrap from the root of mono-repo. We can build the full monorepo from root and then only have it copy files that matter in final image that we publish.

pnpm-workspace.yaml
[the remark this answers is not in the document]
saihaj: I just followed a common structure we keep in our projects. This way `packages` only refers to only standalone packages and `website` is a service that gets deployed so it uses packages.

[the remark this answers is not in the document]
benface: Got it, I guess that makes sense if the `packages` don't get deployed and are all used by the main package!

benface (approved):
Nice. Good work.

PR #299, opened 2023-03-02
ci: cancel previous runs when new commits are pushed

saihaj:
this is to optimize and save long running CI if you are pushing to branch constantly.


benface (approved):
Nice!

PR #300, opened 2023-03-03
Add Mixpanel + migrate to `pnpm` + update dependencies (incl. React 18 and Next 13)

benface:


.env.example
```diff
@@ -1,4 +0,0 @@
-ALGOLIA_API_KEY=
-ALGOLIA_APP_ID=
-NPM_TOKEN=
-SITE_URL=https://thegraph.com/docs
```
benface, line 4: This file was redundant because we already have `.env` checked in the repo. It's weird but [that's what Next recommends](https://nextjs.org/docs/basic-features/environment-variables#default-environment-variables). The `.env` file basically serves as `.env.example` but with some default values (the ones that aren't secrets), and it can be duplicated into `.env.local` to add secrets.

next.config.mjs
```diff
@@ -9,6 +9,8 @@ const env = {
   BASE_PATH: process.env.NODE_ENV === 'production' ? '/docs' : '',
   ALGOLIA_API_KEY: process.env.ALGOLIA_API_KEY,
   ALGOLIA_APP_ID: process.env.ALGOLIA_APP_ID,
+  MIXPANEL_TOKEN:
+    process.env.ENVIRONMENT === 'production' ? 'cfeac8baf33c9b4d255f28d57f3c9148' : 'e57a9892339b2acfd02943c86b746d32',
```
benface, line 13: These are not secrets (that token is included in the frontend bundled).

saihaj (approved):

PR #306, opened 2023-03-08
fixed pipeline credentials

rotarur:


.github/workflows/ci-cd-production.yml
```diff
@@ -26,8 +26,8 @@ jobs:
         uses: docker/login-action@v1
         with:
           registry: ghcr.io
-          username: nenadjaja
-          password: ${{ secrets.GH_ACCESS_TOKEN }}
+          username: ${{ github.actor }}
```
benface, line 29: "actor" 🎭

benface (approved):

.github/workflows/ci-cd-production.yml
rotarur: it's more generic username to use. the important part is the github token that's agnostic

PR #307, opened 2023-03-09
docs: nested sorting

saihaj:
documenting https://github.com/graphprotocol/graph-node/pull/3096

### DIFF
--- a/website/pages/en/querying/graphql-api.mdx
+++ b/website/pages/en/querying/graphql-api.mdx
@@ -49,6 +49,23 @@ When querying a collection, the `orderBy` parameter may be used to sort by a spe
 }
 ```
 
+#### Example for nested entity sorting
+
+Sorting on the basis of nested entities is useful if you are looking to sort entities based on child fields.
+
+In the following example, we sort the tokens by the name of their owner:
+
+```graphql
+{
+  tokens(orderBy: owner__name, orderDirection: asc) {
+    id
+    owner {
+      name
+    }
+  }
+}
+```
+
 ### Pagination
 
 When querying a collection, the `first` parameter can be used to paginate from the beginning of the collection. It is worth noting that the default sort order is by ID in ascending alphanumeric order, not by creation time.
--- a/website/pages/en/querying/graphql-api.mdx
+++ b/website/pages/en/querying/graphql-api.mdx
@@ -51,7 +51,7 @@ When querying a collection, the `orderBy` parameter may be used to sort by a spe
 
 #### Example for nested entity sorting
 
-Sorting on the basis of nested entities is useful if you are looking to sort entities based on child fields.
+As of Graph Node [`v0.30.0`](https://github.com/graphprotocol/graph-node/releases/tag/v0.30.0) entities can be sorted on the basis of nested entities.
 
 In the following example, we sort the tokens by the name of their owner:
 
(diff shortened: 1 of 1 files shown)


website/pages/en/querying/graphql-api.mdx
```diff
@@ -49,6 +49,23 @@ When querying a collection, the `orderBy` parameter may be used to sort by a spe
...
+{
+  tokens(orderBy: owner__name, orderDirection: asc) {
+    id
+    owner {
+      name
+    }
+  }
+}
+```
+
```
azf20, line 68: Is it worth adding a note on constraints (e.g. only works one level deep, doesn't work for interfaces - anything else?)

azf20: presumably it doesn't work if the nested entities are arrays (rather than single linked entities)?

azf20: Presumably it _does_ work on derived entities? I think worth adding these kind of clarifiers

website/pages/en/querying/graphql-api.mdx
```diff
@@ -49,6 +49,23 @@ When querying a collection, the `orderBy` parameter may be used to sort by a spe
 }
 ```
 
+#### Example for nested entity sorting
+
+Sorting on the basis of nested entities is useful if you are looking to sort entities based on child fields.
```
azf20, line 54: This is a bit of a circular sentence - I might make it more of a statement of fact:
"As of Graph Node version 0.30.0, entities can be sorted by fields on nested entities. In the following example..."

azf20 (changes requested):
Nice! Requesting a couple more clarifications and I think will be good to go.

website/pages/en/querying/graphql-api.mdx
```diff
@@ -49,6 +49,25 @@ When querying a collection, the `orderBy` parameter may be used to sort by a spe
...
+  tokens(orderBy: owner__name, orderDirection: asc) {
+    id
+    owner {
+      name
+    }
+  }
+}
+```
+
+> Currently you can sort by one-level deep `String` or `ID` types on `@entity` and `@derivedFrom` fields. Unfortunately, [sorting by interfaces on one level-deep entities](https://github.com/graphprotocol/graph-node/pull/4058) and sorting by nested entities is not supported yet.
```
azf20, line 69: > sorting by nested entities is not supported yet
Do you mean arrays?

azf20 (changes requested):
LGTM - one question on the last sentence

website/pages/en/querying/graphql-api.mdx
saihaj: both 
CleanShot 2023-03-09 at 15 55 42

azf20: I see what you mean - I think worth being very specific then, i.e. "sorting by fields which are arrays, or which themselves are nested entities is not supported". I wonder why we can't use the ID of the `AnotherEntity` to sort, as that will just be a string (except for derived fields)?

saihaj: > I wonder why we can't use the ID of the AnotherEntity to sort, as that will just be a string (except for derived fields)?

I think then it makes it go deeper in the tree. Technically it is just recursively going down the selection set. But we are limiting to 1 level cause of performance reasons.

dotansimha (approved):

azf20 (approved):

PR #309, opened 2023-03-10
ci: next bundle analysis

saihaj:
So we can analyze impacts to bundle size on PRs


website/package.json
```diff
@@ -49,6 +49,7 @@
     "acorn-jsx": "^5.3.2",
     "eslint": "^8.35.0",
     "fast-xml-parser": "^4.1.3",
+    "nextjs-bundle-analysis": "^0.4.0",
```
B2o5T, line 52: you don't need this dependency to be installed

dimaMachina (approved):

PR #314, opened 2023-03-11
use Nextra's `_meta.js` for declaring sidebar titles and order

dimaMachina:
related https://github.com/graphprotocol/docs/issues/311

- [x] refactor to use `_meta.js`
- [x] remove injected `getStaticProps` for each markdown page


Dockerfile
```diff
@@ -3,7 +3,7 @@ FROM node:18-alpine as builder
 ENV PNPM_HOME="/usr/bin"
 
 RUN apk add --no-cache git
-RUN npm install -g pnpm
+RUN npm install -g pnpm@7
```
benface, line 6: Should we be even more specific and install version 7.28.0, as that's the one set in `package.json`? (Or can we be less specific in `package.json`?)

package.json
```diff
@@ -7,8 +7,8 @@
     "build": "pnpm -r build",
     "start": "pnpm --filter @graphprotocol/docs start",
     "export": "pnpm --filter @graphprotocol/docs export",
-    "lint": "pnpm typecheck && eslint . --ext .js,.jsx,.ts,.tsx,.mjs --cache --max-warnings 0 && pnpm prettier:check",
-    "lint:fix": "pnpm typecheck; eslint . --ext .js,.jsx,.ts,.tsx,.mjs --fix; pnpm prettier; exit 0",
+    "lint": "eslint --cache --ext .js,.jsx,.ts,.tsx,.mjs --max-warnings 0 . && pnpm prettier:check && pnpm typecheck",
+    "lint:fix": "eslint --cache --ext .js,.jsx,.ts,.tsx,.mjs --fix . && pnpm prettier && pnpm typecheck",
```
benface, line 11: Oh, why revert this?

website/components/index-page.tsx
benface, line 1: Why the kebab-case filename? Could we not call it `IndexPage.tsx`?

Dockerfile
saihaj: the package json one has no impact here. We probably want to upgrade to `pnpm@8` but that should be done independently.

website/components/index-page.tsx
B2o5T: renamed

package.json
B2o5T: sorry, mistake from rebase, now replaced only `;` by `&&` as I mentioned https://github.com/graphprotocol/docs/pull/319#discussion_r1140659507

saihaj (approved):

Dockerfile
benface: I guess I'm just saying there's an inconsistency where `package.json` says "you're supposed to install these dependencies with pnpm v7.28.0" (right?) and Dockerfile will actually install them with pnpm v7.30.5 because that's the latest v7.x.x. But maybe I'm missing the point of the `packageManager` key.

website/pages/es/cookbook/migrating-a-subgraph.mdx
```diff
@@ -215,5 +215,5 @@ Si todavía estás confundido, ¡no temas! Consulta los siguientes recursos o mi
 
 - [Los Contratos de The Graph Network](https://github.com/graphprotocol/contracts)
 - [Contrato de Curación](https://github.com/graphprotocol/contracts/blob/dev/contracts/curation/Curation.sol) - el contrato subyacente en el que se envuelve el GNS
-  - Address - `0x8fe00a685bcb3b2cc296ff6ffeab10aca4ce1538`
+- Address - `0x8fe00a685bcb3b2cc296ff6ffeab10aca4ce1538`
```
benface, line 218: Not sure why this changed, but:
```suggestion
  - Address - `0x8fe00a685bcb3b2cc296ff6ffeab10aca4ce1538`
```

benface (approved):
Really good work. 🙌

website/pages/es/cookbook/migrating-a-subgraph.mdx
B2o5T: you are right

Dockerfile
B2o5T: I think you have a point @benface ! I am +1 for consistency with package.json

B2o5T:
> Really good work. 🙌

Thank you!

the only thing that it's annoying to recreate [`_meta.json`/`_meta.js`](https://nextra.site/docs/guide/organize-files#_metajson) for each language

maybe I could rethink of way to have 1 folder that will contain `_meta.js` files and other languages will extend it

benface:
> maybe I could rethink of way to have 1 folder that will contain _meta.js files and other languages will extend it
As the solution to put them to [...locale]

Oh, that's a nice idea! It would still be possible to override a given `_meta.js` file in a given locale, right?

B2o5T:
> Oh, that's a nice idea! It would still be possible to override a given _meta.js file in a given locale, right?

Yes, `_meta.js` in language directory should have higher priority than meta in catch all route `[...locale]`. I can do it in separate PR and update it in Nextra

PR #315, opened 2023-03-14
Suggest the use of "Token ID" to deprecate the subgraph via Etherscan

brunowowk:
The 'SubgraphId' is presented in a base58 representation in the dashboards, which is not accepted on Etherscan uint256 parameters.

The Token Id, on the other hand, uses the decimal representation, which is accepted on Etherscan.


Web3Slimchance (approved):
Great catch! Thank you.

PR #301, opened 2023-03-08
The Graph Docs x Nextra

dimaMachina:
<s>I'll rebase to main once `saihaj/move-to-monorepo` will be merged</s>

Follow up items
- https://github.com/graphprotocol/docs/issues/310
- https://github.com/graphprotocol/docs/issues/311

PR #319, opened 2023-03-16
Benface/Nextra

benface:
Some proposed changes to accompany my review of #301...


saihaj:
@benface if you can rebase from `nextra` branch and just have your change commits cause right now the diff is noise cause of the changes from `main` which were not synced yet

benface:
@saihaj – I had intentionally merged `main` into my branch, but I guess that's problematic, especially with squashing commits. I just did like you said. Better now?

saihaj:
> @saihaj – I had intentionally merged `main` into my branch, but I guess that's problematic, especially with squashing commits. I just did like you said. Better now?

yeah now we can easily go through commits!

website/components/EditPageLink.tsx
```diff
@@ -16,12 +17,12 @@ export const EditPageLink = ({ mobile = false, ...props }: EditPageLinkProps) =>
...
   const { pagePath: _pagePath } = useContext(NavContext)!
   const pagePathSegments = _pagePath.split('/')
-  pagePathSegments[0] = ['en', '[locale]'].includes(pagePathSegments[0]) ? pagePathSegments[0] : 'en'
+  pagePathSegments[1] = ['en', '[locale]'].includes(pagePathSegments[1]) ? pagePathSegments[1] : 'en'
   const pagePath = pagePathSegments.join('/')
 
   return (
     <Link
-      href={`https://github.com/graphprotocol/docs/blob/main/pages/${pagePath}`}
+      href={`https://github.com/graphprotocol/docs/blob/main/website/${pagePath}`}
```
B2o5T, line 25: nice catch for edit on github, forget to test it

website/tsconfig.json
```diff
@@ -1,11 +1,4 @@
 {
   "extends": "../tsconfig.json",
-  "compilerOptions": {
-    "baseUrl": ".",
-    "paths": {
-      "@/*": ["*"]
```
saihaj, line 6: I think this was nice :)

website/next.config.mjs
```diff
@@ -20,13 +20,14 @@ const withNextra = nextra({
   },
   transform(content, { route }) {
     if (route) {
+      const routeSegments = route.split('/')
       return `
-import { getNavItems } from '@/navigation'
+import { getNavItems } from '${'../'.repeat(routeSegments.length - 1)}navigation'
```
B2o5T, line 25: why did you remove `@` alias in tsconfig? seems it worked well?

benface: It didn't! Try running `pnpm lint` in your branch.

dimaMachina (approved):

website/tsconfig.json
benface: I agree, if you find a way to make it pass `pnpm lint`, let's put it back!

website/next.config.mjs
saihaj: I think ESlint config has a way to pick aliases 🤔

saihaj (approved):

website/next.config.mjs
B2o5T: image

benface: Oh yeah Prettier fails which prevents `tsc` from even running... try `pnpm lint:fix`

benface: We should probably replace the `&&`s with `;` so each command runs regardless of the result of the previous one.

B2o5T: to fix you can replace the below in `package.json`

```diff
-"typecheck": "tsc --noEmit",
+"typecheck": "pnpm --filter @graphprotocol/docs typecheck",
```

benface: Ohhh, thank you @B2o5T! Done in https://github.com/graphprotocol/docs/pull/319/commits/a0717fa035aa6a0ed4fc4524bf737a0e270db4d9, and then I reverted removing the alias in https://github.com/graphprotocol/docs/pull/319/commits/e8baaf649f0933da1b572dfba25ce2fe0aab29fe.

By the way, there are still 4 ESLint warnings:

CleanShot 2023-03-16 at 22 20 47@2x

Those are weird, I don't really understand why TypeScript thinks they are `any`.

B2o5T: I will take a look tomorrow!

package.json
```diff
@@ -7,11 +7,11 @@
     "build": "pnpm -r build",
     "start": "pnpm --filter @graphprotocol/docs start",
     "export": "pnpm --filter @graphprotocol/docs export",
-    "lint": "eslint . --ext .js,.jsx,.ts,.tsx,.mjs && pnpm prettier:check && pnpm typecheck",
-    "lint:fix": "eslint . --fix --ext .js,.jsx,.ts,.tsx,.mjs && pnpm prettier && pnpm typecheck",
+    "lint": "eslint . --ext .js,.jsx,.ts,.tsx,.mjs; pnpm prettier:check; pnpm typecheck",
+    "lint:fix": "eslint . --fix --ext .js,.jsx,.ts,.tsx,.mjs; pnpm prettier; pnpm typecheck",
```
B2o5T, line 11: @benface we should use `&&` instead, otherwise, it's not fails if some command fails, e.g. `eslint --max-warnings 0`

benface: Ah yeah, oops. I guess I just find it annoying that if e.g. Prettier fails, the TS errors are not even shown. Maybe Prettier should be last?

B2o5T: yes, I guess the following commands `typecheck` -> `eslint` -> `prettier` will be better

PR #333, opened 2023-03-27
Avoid using `h1` in page content

benface:
There should be a single `h1` per page, and it's the page's title (automatically rendered by the template, so we should never have a single `#` in an MDX file).

cc @MichaelMacaulay


dimaMachina (approved):

benface:
Oh that's a great idea @B2o5T! Is there an existing rule for it?

B2o5T:
I think I already saw something similar

Update: https://www.npmjs.com/package/remark-lint-first-heading-level

benface:
Thank you @B2o5T, that would be awesome. I tried doing it in [a new branch named `benface/fix-headings-with-lint`](https://github.com/graphprotocol/docs/commits/benface/fix-headings-with-lint) but I don't see any warning or error when I run `pnpm build`...

PR #334, opened 2023-03-27
Forbid h1 `# ...` elements, enforce h2 `## ...` as first heading on page

dimaMachina:


.eslintrc.cjs
```diff
@@ -0,0 +1,27 @@
...
+      },
+    },
+    {
+      files: ['*.{ts,tsx}'],
+      parserOptions: {
+        project: 'tsconfig.json',
+      },
+    },
+    {
+      // We lint only english pages because other languages will be translated from english
```
benface, line 17: 👍

website/pages/en/cookbook/subgraph-uncrashable.mdx
```diff
@@ -4,7 +4,7 @@ title: Safe Subgraph Code Generator
 
 [Subgraph Uncrashable](https://float-capital.github.io/float-subgraph-uncrashable/) is a code generation tool that generates a set of helper functions from the graphql schema of a project. It ensures that all interactions with entities in your subgraph are completely safe and consistent.
 
-### Why integrate with Subgraph Uncrashable?
+## Why integrate with Subgraph Uncrashable?
```
benface, line 7: Was this rewritten with `pnpm lint:fix`, or did you change it manually? If the former, what rule is responsible for it? I'm not sure we always want to follow h2 => h3 => h4.

B2o5T: I changed it manually, remark lint doesn’t support auto fix

B2o5T: > I'm not sure we always want to follow h2 => h3 => h4.

it's a good practice, especially for TOC, so every heading will have right amount of `margin-left`

image

B2o5T: also there is a rule for this case that we could also extend https://github.com/remarkjs/remark-lint/tree/main/packages/remark-lint-heading-increment

benface: Hmm, so I would tend to be a bit less strict here and only keep the custom rule that you created (`remark-lint-restrict-elements` to forbid `h1` altogether). The `remark-lint-first-heading-level` rule is a bit strange: it will error if the first heading is an `h3` instead of an `h2`, but it won't care at all if an `h2` is immediately followed by an `h5` instead of an `h3`. I guess that's what `remark-lint-heading-increment` is for, but I feel like we should allow for exceptions and leave it up to code reviewers. For instance, for FAQs, I find that smaller questions look better (and I'm not bothered by the subtle margin difference in the TOC that you mentioned):

_Before_

CleanShot 2023-03-27 at 22 31 17@2x

_After_

CleanShot 2023-03-27 at 22 31 25@2x

I would like someone else to chime on this though. I could easily be convinced.

B2o5T: >  I guess that's what remark-lint-heading-increment is for,

exactly

> I would like someone else to chime on this though. I could easily be convinced.

sure, let's hear through others and anyway I could do a rollback of changes

benface: Cool, tagging @MichaelMacaulay since he contributes the most to actual content, and @Web3Slimchance, @azf20, @nenadjaja, and @schmidsi because they might have an opinion. The question is: do we want to allow skipping heading levels (other than `h1` which is always there at the top), e.g.  having `h3`s without an `h2`, like in the first FAQs screenshot I posted above?

benface: @B2o5T – Michael and Adam are OK with `remark-lint-first-heading-level` + `remark-lint-heading-increment`. Let's keep them and merge this when you can!

benface (approved):
Great work, thank you!

benface (commented):
@B2o5T – Actually, I just realized that this PR does _not_ currently include [`remark-lint-heading-increment`](https://github.com/remarkjs/remark-lint/tree/main/packages/remark-lint-heading-increment). We can still merge it as is, but if you want to add it first, that'd be great.

B2o5T:
@benface let's add [remark-lint-heading-increment](https://github.com/remarkjs/remark-lint/tree/main/packages/remark-lint-heading-increment?rgh-link-date=2023-04-25T23%3A38%3A27Z) in another PR!

PR #336, opened 2023-03-31
extract theme as separate package `@graphprotocol/theme`

dimaMachina:


packages/theme/package.json
```diff
@@ -0,0 +1,58 @@
+{
+  "name": "@graphprotocol/theme",
```
saihaj, line 2:
```suggestion
  "name": "@graphprotocol/nextra-theme",
```

B2o5T: yes! `nextra-theme` sounds better

dotansimha (approved):

saihaj (approved):

packages/theme/package.json
B2o5T: done!

packages/theme/src/components/Link.tsx
```diff
@@ -1,14 +1,12 @@
...
 
-import { buildShadow, buildTransition, useTheme } from '@edgeandnode/components'
-
-import { AppLocale, useI18n } from '@/i18n'
+import { buildShadow, buildTransition, useI18n, useTheme } from '@edgeandnode/components'
 
 export type LinkProps = Pick<NextLinkProps, 'replace' | 'scroll' | 'shallow' | 'prefetch'> &
   Omit<AnchorHTMLAttributes<HTMLAnchorElement>, 'href'> & {
     href?: NextLinkProps['href']
-    locale?: AppLocale
```
benface, line 11: Why remove all references to `@/i18n`?

benface: If anything, I think `@/i18n` should be part of the theme, and `website` should import it from the theme.

website/components/_app.tsx
```diff
@@ -61,6 +64,50 @@ const DefaultSeoWithLocale = () => {
...
+    const pageMap = (__nextra_internal__.pageMap as PageMapItem[]).find(
+      (pageItem): pageItem is Folder => pageItem.kind === 'Folder' && pageItem.name === locale
+    )!.children
+    pageMap.find((pageItem): pageItem is MetaJsonFile => pageItem.kind === 'Meta')!.data.index = translate(
+      translations,
+      locale,
+      'index.title'
+    )
+    return pageMap
+  }, [__nextra_internal__.pageMap, hasMDXPage, locale])
```
benface, line 110: This whole thing seems pretty hacky. 😅

website/package.json
```diff
@@ -1,33 +1,31 @@
...
     "start": "next start",
     "typecheck": "tsc --noEmit"
   },
   "dependencies": {
-    "@docsearch/react": "^3.3.3",
     "@edgeandnode/components": "^27.4.1",
     "@emotion/react": "^11.10.6",
-    "@radix-ui/react-collapsible": "1.0.2",
-    "@radix-ui/react-visually-hidden": "^1.0.2",
+    "@graphprotocol/nextra-theme": "workspace:*",
```
benface, line 20: Nice, I didn't know that trick.

benface (approved):
I left comments for potential improvements, but this is great, thank you!

packages/theme/src/components/Difficulty.tsx
```diff
@@ -1,8 +1,8 @@
-import { useI18n } from '@/i18n'
+import { useI18n } from '@edgeandnode/components'
```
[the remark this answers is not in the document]
B2o5T: this function should be generic (to accept different translations) since now `useI18n` lives in `packages/theme` (or maybe I should rename it `packages/nextra-theme`)

A current function definition is

```ts
export const useI18n = () => _useI18n<typeof translations>()
```

packages/theme/src/components/Link.tsx
B2o5T: I described https://github.com/graphprotocol/docs/pull/336#discussion_r1169369867 why I removed all `@/i18n` usages

website/package.json
B2o5T: it assume to use local version

website/components/_app.tsx
B2o5T: it is hacky due `[locale]/index.mdx` file if we'll refactor to `en/index.mdx`, `fr/index.mdx` the whole mentioned code will be a lot simplified (because normally TOC headings are parsed but here I inject these headings only for this `index` page)

packages/theme/src/components/Difficulty.tsx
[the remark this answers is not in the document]
benface: Oh, right, each package that uses the theme might have its own translatable strings... still, the theme itself will only ever refer to strings that exist within the theme, so it's almost like we'd need 3 layers of this. Oh well, I guess this works, it just doesn't provide the nice autocomplete experience. :P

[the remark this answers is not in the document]
B2o5T: yeah, we lose autocomplete

website/components/_app.tsx
B2o5T: custom styles can be extracted as React components but headings need to exist as raw markdown headings

```md
---
title: My title
---

## Foo bar

<SomeSections />

### Qux

<SomeLogos />

```

benface: Ok, good idea.

B2o5T: I will create an issue for 👆, will rename `theme` folder to `nextra-theme` and we are ready to merge it!

The next one is Remote MDX PR!

PR #352, opened 2023-04-26
Update creating-a-subgraph.mdx

Tonio2:
Fix typo in writing mappings code example


schmidsi:
@Tonio2 Thanks for the PR. Recently there was a feature introduced: "Bytes as IDs". So the ID does not need to be a string. That being said, are you running into troubles with that code example? I'm not 100% if the event param is not a BigInt that needs to be converted to Bytes first.

Read more about it here: https://medium.com/edge-node-engineering/two-simple-subgraph-performance-improvements-a76c6b3e7eac

0xa3k5:
Thank you @Tonio2  for your contribution and for taking the time to improve the docs. Closing the PR due to inactivity. Please feel free to reopen it or create a new one if you find it necessary.

PR #365, opened 2023-05-01
Added sorting cookbook page

marcusrein:
Sorting cookbook draft PR added as per Simon.

### DIFF
--- a/website/pages/en/cookbook/sorting.mdx
+++ b/website/pages/en/cookbook/sorting.mdx
@@ -1,0 +1,172 @@
+---
+title: GraphQL Sorting Strategies
+---
+
+## Sorting Entity IDs
+
+When creating subgraphs in The Graph, it's important to consider the type of ID used for your entities. This choice can have a significant impact on the performance and ease of use of your subgraph.
+
+If accurate sorting is desired, both Strings and Bytes for IDs are not optimal as the sorting is done non-sequentially which can lead to unexpected results. If attempting to sort by id of type "ID" (which is a string), the ordered sequence might look like this:
+
+schema.graphql:
+
+```graphql
+type Transfer @entity {
+  id: ID!
+  from: Bytes! # address
+  to: Bytes! # address
+  tokenId: BigInt! # uint256
+}
+```
+
+Query:
+
+```graphql
+{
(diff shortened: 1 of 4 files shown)


website/pages/en/cookbook/sorting.mdx
```diff
@@ -0,0 +1,172 @@
+---
+title: GraphQL Sorting Strategies
+---
+
+## Sorting Entity IDs
+
+When creating subgraphs in The Graph, it's important to consider the type of ID used for your entities. This choice can have a significant impact on the performance and ease of use of your subgraph.
```
schmidsi, line 7: This is actually no a big consideration. As of now, IDs should always be `Bytes`.

schmidsi: Unless there is a good reason to use `ID` which translates to string. Examples might be traits or unique names.

website/pages/en/cookbook/sorting.mdx
```diff
@@ -0,0 +1,172 @@
...
+
+If accurate sorting is desired, both Strings and Bytes for IDs are not optimal as the sorting is done non-sequentially which can lead to unexpected results. If attempting to sort by id of type "ID" (which is a string), the ordered sequence might look like this:
```
schmidsi, line 9: What is accurate? Only for numerical values, sorting by `Bytes` and `String` feels off. If the ID is a transactionHash or a string, that sorting is fine.

So I suggest to be clear here.

website/pages/en/cookbook/sorting.mdx
```diff
@@ -0,0 +1,172 @@
...
+      {
+        "id": "0x01020000",
+        "tokenId": "513"
+      }
+    ]
+  }
+}
+```
+
+## Ideal Sorting Strategy of IDs
```
schmidsi, line 119: Again, this really depends on the use-case. So ideal for numerical values is not ideal for strings or bytes.

schmidsi (changes requested):
Generally it is a very good cook-book topic. Although I think we should be very specific about the details and not only talk about the numerical value use-case.

azf20:
@marcusrein @schmidsi checking on this one?

PR #366, opened 2023-05-01
docs: load in block

saihaj:
Adding docs for https://github.com/graphprotocol/graph-tooling/issues/1305


website/pages/en/developing/assemblyscript-api.mdx
```diff
@@ -280,6 +280,20 @@ As the entity may not exist in the store yet, the `load` method returns a value
 
 > **Note:** Loading entities is only necessary if the changes made in the mapping depend on the previous data of an entity. See the next section for the two ways of updating existing entities.
 
+#### Looking up entities created withing a block
```
saihaj, line 283: should I also add the version of `graph-node` and `graph-cli` this feature is available?

dotansimha: I think it good to mention this somewhere

azf20: Agree, something like "This helper is generated by `graph-cli` for all entities as of v0.x.x, and subgraphs using this functionality can be deployed to Graph Node from v0.31.0."

website/pages/en/developing/assemblyscript-api.mdx
```diff
@@ -280,6 +280,20 @@ As the entity may not exist in the store yet, the `load` method returns a value
...
+
+The store API facilitates the retrieval of entities that were created or updated in the current block. This feature proves to be useful in situations where the handler requires the on-chain event to be generated in the same block. By utilizing this functionality, indexing performance can improve significantly.
```
azf20, line 285: > This feature proves to be useful in situations where the handler requires the on-chain event to be generated in the same block

I think here we can borrow from @lutter's description from the graph-node PR...

A typical situation for this is that one handler creates a Transaction from some on-chain event, and a later handler wants to access this transaction if it exists. In the case where the transaction does not exist, the subgraph will have to go to the database just to find out that the entity does not exist; if the subgraph author already knows that the entity must have been created in the same block, using `loadInBlock` avoids this database roundtrip. For some subgraphs, these missed lookups can contribute significantly to the indexing time.

azf20: I think also worth noting that if there is no entity created "in block" (even if there **is** an entity with the right ID in the database), this will return `null`

azf20 (changes requested):
Thanks so much! Some small comments

azf20 (approved):

dotansimha (approved):

PR #367, opened 2023-05-01
fix: point docs to graph tooling for example subgraph

saihaj:
In plans to unify all tooling, one of the things we did last week was migrate all the example subgraphs to `graph-tooling` https://github.com/graphprotocol/graph-tooling/pull/1204 and we are going to archive the `example-subgraphs` repo https://github.com/graphprotocol/example-subgraphs/pull/5. So merging this docs is the last step for us to be able to archive the old example subgraph repo.


dotansimha (approved):

azf20:
I think this LGTM - just checking with @benface that changing all the non "en" language pages will be ok with Crowdin?

benface:
@azf20 – Sorry for the delay. It will be "ok" in the sense that it'll work for now, but unless translators apply the same changes in Crowdin (which they should because changing the source string will mark the translated strings as outdated, but it might take some time), they will be overwritten the next time we merge the Crowdin PR.

PR #372, opened 2023-05-05
move substreams docs to own sidebar section

dimaMachina:
https://guild-oss.slack.com/archives/C0210QKF9HT/p1683284179208799


website/pages/ko/_meta.js
```diff
@@ -1,5 +1,5 @@
 import meta from '../en/_meta.js'
 
 export default {
-  ...meta,
+  ...JSON.parse(JSON.stringify(meta)),
```
saihaj, line 4: is meta not already JS object?

B2o5T: it's for copy object, Nextra enhance title and since we use same `_meta` in different language (for sidebar order) we can have a bug to have another translation in the sidebar

website/next.config.js
```diff
@@ -17,7 +17,7 @@ const withNextra = nextra({
   transform(result, { route }) {
     if (route && !result.includes('getStaticProps')) {
       const banner = `
-import { getPageMap } from '@/components/get-page-map'
+import { getPageMap } from '@/src/get-page-map'
```
B2o5T, line 20: since now `components` folder contain `utils` files `get-page-map` I renamed this folder to just `src`

website/pages/en/_meta.js
```diff
@@ -27,11 +27,17 @@ export default {
     type: 'separator',
   },
   '###2': {
+    type: 'heading',
+    title: 'Substreams',
+  },
+  substreams: {
+    type: 'children'
```
B2o5T, line 34: added new prop to theme to not render `folder` element and render directly his files (sidebar links for substreams docs in our case)

website/next.config.js
benface: @B2o5T – Why the kebab case? I feel like it should be `getPageMap.ts`.

website/pages/ar/_meta.js
```diff
@@ -1,5 +1,5 @@
 import meta from '../en/_meta.js'
 
 export default {
-  ...meta,
+  ...JSON.parse(JSON.stringify(meta)),
```
benface, line 4: Could we use [`structuredClone()`](https://developer.mozilla.org/en-US/docs/Web/API/structuredClone) instead?

benface (approved):
Nice, simple implementation. 👍

benface (approved):
Nice, simple implementation. 👍

website/pages/ko/_meta.js
benface: Could we use [structuredClone()](https://developer.mozilla.org/en-US/docs/Web/API/structuredClone) instead?

PR #385, opened 2023-05-24
Update docs for indexer setup

fordN:
The docker-compose setup is designed specifically for graph-node devs, so I'm removing reference of it from the docs for network indexers. The removed section can likely be replaced later with a section on using the [stakesquid docker-compose setup](https://github.com/StakeSquid/graphprotocol-mainnet-docker) which was designed for use by indexers and includes the other necessary components of the indexer stack.


azf20 (approved):
Thanks!

saihaj (approved):

PR #405, opened 2023-06-07
Turn Supported Networks into a table

benface:
CleanShot 2023-06-09 at 16 41 09@2x


website/pages/en/developing/supported-networks.mdx
```diff
@@ -4,35 +4,37 @@ title: Supported Networks
...
-The following networks are supported on The Graph's Decentralized Network:
-
-<ListUnordered>
-  {getSupportedNetworks('studio').map((network) => (
-    <li key={network.cliName}>
-      <CodeInline>{network.cliName}</CodeInline>
-      {network.cliName === 'mainnet' ? ` (${network.name})` : null}
-      {network.beta ? '*' : null}
-    </li>
+Here is a table showing the networks that The Graph supports, whether it is on the hosted service, Subgraph Studio, or the decentralized network:
```
benface, line 7: Not sure about this text. Aren't the rows actually "chains" rather than "networks"? Should we rename this entire page?

cmwhited (approved):

website/pages/en/developing/supported-networks.mdx
azf20: Agree about the text - I think the table is self-explanatory, so we can probably drop it?
On chains vs. networks, that is a general naming issue we have, but I am ok with keeping network here as `network` is what it is called in the subgraph manifest

azf20 (approved):
I like it!

azf20:
One thing, and I think it's to do with the source data rather than this PR, but I think the labelling of Fantom and Fantom Opera is incorrect? i.e. Fantom Opera is the testnet
image

website/pages/en/developing/supported-networks.mdx
```diff
@@ -4,35 +4,37 @@ title: Supported Networks
...
+    <th align="center">Chain ID</th>
+    <th align="center">Hosted Service</th>
+    <th align="center">Subgraph Studio</th>
+    <th align="center">Decentralized Network</th>
+  </tr>
+  {getSupportedNetworks().map((network) => (
+    <tr key={network.cliName}>
+      <td>
+        {network.name}
+        {network.isBeta ? '*' : null}
```
azf20, line 22: I think we don't mention what the asterisks mean anywhere? I think this is principally relevant on the network, so maybe the asterisks should just show there (for chains which are on the decentralised network, next to the check mark?)

azf20 (changes requested):
One point around the "beta" asterisks, and I think we may need a fix in Common for Fantom labelling https://github.com/edgeandnode/common/blob/main/src/constants/network.ts#L258

benface:
@azf20 –

> One thing, and I think it's to do with the source data rather than this PR, but I think the labelling of Fantom and Fantom Opera is incorrect? i.e. Fantom Opera is the testnet

[Is it?](https://www.fantom.foundation/what-is-fantom-opera/) cc @cmwhited 

---

> I think we don't mention what the asterisks mean anywhere?

Sorry, I didn't include it in my screenshot, but we do mention it right below the table:

CleanShot 2023-06-08 at 09 18 07@2x

--- 
> I think this is principally relevant on the network, so maybe the asterisks should just show there (for chains which are on the decentralised network, next to the check mark?)

Done in https://github.com/graphprotocol/docs/pull/405/commits/94c8f0f04996747a70e04361f5169ee5d7ec953c

website/pages/en/developing/supported-networks.mdx
benface: Thank you! I removed the paragraph in https://github.com/graphprotocol/docs/pull/405/commits/94c8f0f04996747a70e04361f5169ee5d7ec953c.

MichaelMacaulay (commented):
Thanks so much for putting this together! I only have 2 minor comments:

1. I believe Ethereum Rinkeby is deprecated but still seems to be showing in the table?
2. Do we need to include the chain ID? I don't think its a big value add for our docs & removing that one column will make a big difference for simplifying this chart imo

benface:
@azf20 & @MichaelMacaulay – I guess I should just not render any network that doesn't have any check in the table, right? So that would hide Rinkeby, Optimism Kovan, and Arbitrum Rinkeby.

benface:
Done @azf20 & @MichaelMacaulay. I updated the screenshot on the first post. 👆

Want to give this one last review? 🙏

azf20 (approved):

PR #407, opened 2023-06-08
setup turbo for `dev` and `build` commands

dimaMachina:
fixes https://guild-oss.slack.com/archives/C0210QKF9HT/p1686132196894879

PR #411, opened 2023-06-11
update README.md

itsjerryokolo:
It would be helpful to state that `pnpm build` needs to be run first before running `pnpm dev`
The page doesn't load if you don't build first. Took me a while to figure it out.


B2o5T:
https://github.com/graphprotocol/docs/pull/407 will fix this issue

PR #412, opened 2023-06-12
Moving Graph Node and MIPs FAQ

MichaelMacaulay:


pnpm-lock.yaml
benface, line 1: @MichaelMacaulay – I think you accidentally deleted this file?

MichaelMacaulay: @benface fixed this! could you take another look when you get a chance?

website/pages/en/_meta.js
```diff
@@ -42,6 +42,8 @@ export default {
   },
   'operating-graph-node': '',
   firehose: 'Firehose',
+  graphcast: 'Graphcast',
+  'mips-faqs': 'MIPs FAQs',
```
benface, line 46: Since you're not overriding the titles, better to set them to an empty string to use what's set in the frontmatter of the respective MDX files. That way, if it changes, we only have to update it in one place.
```suggestion
  graphcast: '',
  'mips-faqs': '',
```

MichaelMacaulay: Ahh, makes sense. Thank you!

benface (approved):
👍

PR #414, opened 2023-06-13
Add 302 redirect from `/en/firehose/` to `/en/firehose/README/`

benface:
Like #378 but for the Firehose docs, this prevents a 403 when switching from a non-English language to English while on the Firehose page.


nginx.conf
```diff
@@ -47,7 +47,6 @@ http {
         # Permanent redirects (301)
         rewrite ^/docs/hostedservice/(.*)$                              $scheme://$http_host/docs/en/hosted-service/$1      permanent;
         rewrite ^/docs/(?!(?:[a-zA-Z][a-zA-Z]|_next|img)(?:/|$))(.*)$   $scheme://$http_host/docs/en/$1                     permanent; # Redirect to `/en` if no language in URL and not an asset URL
-        rewrite ^/docs/en/substreams/$                                  $scheme://$http_host/docs/en/substreams/README/     permanent;
```
benface, line 50: I also changed this one to a temporary one (302 instead of 301). See below.

saihaj (approved):

PR #431, opened 2023-06-28
add open graph images + multi langs + image snapshots tests

### LINKED ISSUE #402: opengraph image service
https://github.com/the-guild-org/docs/tree/main/packages/og-image

dimaMachina:
closes #402


benface:
Would also be nice (though maybe out of scope for this PR?) to set the `og:description` to a snippet of the page's content, instead of the default description that all pages have ("Browse the latest developer documentation including API reference, articles, and sample code"). Even cooler if we used GPT to write a summary of the page, but that is definitely out of scope, haha.

benface (approved):
Super cool! HTML → SVG → image 😁 Good job!

PR #458, opened 2023-08-03
Fix URL in quick start page

kristians-kuhta:
The previous URL was broken.

This PR fixes it to point to:
https://thegraph.com/docs/en/cookbook/quick-start/


benface:
Thank you for highlighting this issue, @kristians-kuhta. However, the `../querying/querying-the-graph/` link should be correct, as `querying` is just one level below `cookbook/quick-start`. A better fix would be to use an absolute link (e.g. `/querying/querying-the-graph/`), but an even better fix is to make relative links work as expected. 😄 That's what I did in #459. So, I'm going to close this PR, but thanks again!

kristians-kuhta:
OK, I was kind of confused about why one would need to do `../../`, but it did work at the moment.
Thanks!

PR #459, opened 2023-08-04
Fix relative links

benface:
Another PR to fix links. 😅

This one is a better fix for the issue that #458 works around.


saihaj (approved):
wonder if we can have some way to add tests for this 🤔

benface:
> wonder if we can have some way to add tests for this

We definitely should. I think something simple like ensuring there are no 404s (or maybe other error codes too) in the built HTML would be good.

PR #465, opened 2023-08-10
Fix search regression

benface:
This PR fixes a regression in the search feature caused by [the removal of a `span`](https://github.com/graphprotocol/docs/commit/1437d84467e75bf0d127c238f0e59c2349da7339#diff-d3c6a023eff820e26b00e91e1b8a72ab6c77fe1f4efeec696edbdb5d4949d1c2L35), which causes most search results to show "#Link to this section". 😬

CleanShot 2023-08-09 at 22 03 03@2x


algolia-crawler.config.js
benface, line 1: This file is for reference only, but I've made the change on the Algolia Crawler admin.

saihaj:
I still see it if u type one word 


CleanShot 2023-08-09 at 21 08 26

benface:
@saihaj – Unfortunately the fix won't be visible locally, because this PR needs to be deployed and the crawler to be restarted for the index to update. You'll just have to trust that it works. 😅

saihaj (approved):

PR #454, opened 2023-08-01
Update indexing documentation for Exponential Rebates

alex-pakalniskis:
Removes references to Cobb-Douglas, rebate pools, and legacy allocation states. Tagging @tmigone for visibility.

Thanks :smile:

PR #470, opened 2023-08-10
chore: add exponential rebate tldr

alex-pakalniskis:
https://github.com/graphprotocol/docs/pull/454#issuecomment-1672359973

@howard-the-graph please confirm if this looks good to go.

### DIFF
--- a/website/pages/en/network/indexing.mdx
+++ b/website/pages/en/network/indexing.mdx
@@ -81,7 +81,7 @@ Disputes can be viewed in the UI in an Indexer's profile page under the `Dispute
 
 ### What are query fee rebates and when are they distributed?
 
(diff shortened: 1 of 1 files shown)


howard-the-graph (commented):
I recommend a minor revision to be more concise and avoid vague phrases about relative amounts:

> The exponential rebate function is proposed as a way to ensure indexers achieve the best outcome by faithfully serving queries. It works by incentivizing Indexers to allocate stake, which can be slashed for erring when serving a query.

howard-the-graph:
Separately, the phrase "by the gateway and" should be removed from the first sentence of the paragraph.

howard-the-graph (approved):

azf20 (approved):

benface:
@alex-pakalniskis – Uhhh yeah that's weird:

CleanShot 2023-09-12 at 11 20 34@2x

I can't imagine the change in this PR introducing any sort of failure, so I would just ignore it yes.

benface:
cc @saihaj @B2o5T ☝️

PR #483, opened 2023-08-22
Moving quick start and adding redirect

MichaelMacaulay:


benface (changes requested):
Could you apply the changes (moving the file and updating `_meta.js`) to languages other than English as well? 🙏 I don't think Crowdin is smart enough to detect that a file was moved or deleted; it will recreate the translated file in the new location but won't delete the old one.

benface (approved):

PR #486, opened 2023-08-24
Update dependencies + improve ESLint/TS configs + delete orphan pages

benface:
See commits for more details.


package.json
```diff
@@ -5,19 +5,19 @@
   "scripts": {
     "build": "NODE_OPTIONS='--max_old_space_size=4096' turbo run build",
     "check": "pnpm typecheck && pnpm lint && pnpm prettier:check",
-    "check:fix": "pnpm typecheck && pnpm lint:fix",
+    "check:fix": "pnpm typecheck; pnpm lint:fix",
```
B2o5T, line 8: why this change? if `typecheck` fails you still want to run `pnpm lint:fix`? `&&` enforce to stop immediately command and return exit code 1, `;` doesn't return exit code at all

```json
"lint": "eslint --cache --ignore-path .gitignore . && echo hello",
```
image

```json
"lint": "eslint --cache --ignore-path .gitignore . ; echo hello",
```

image

dimaMachina (approved):
lgtm, just curious about `;` change and why some pages were removed, to reindex them by crowdin?

package.json
benface: @B2o5T – My reasoning is that the `check` script should fail _as early as possible_, since it's used in CI to check if everything's good to go or if there's an issue (any issue at all). Whereas the `check:fix` script is meant to be run locally (I personally run it before every commit), and it should list _all_ the issues and fix as many of them as possible. For instance, I don't feel that an ESLint error should be a reason not to format the code using Prettier. And since we only run it locally, `check:fix` doesn't need to return an error code. I suppose it wouldn't hurt if it did, but I'm not sure if there's an easy way to do that while not stopping early if the first process returns a non-zero exit code?

benface:
> just curious about ; change and why some pages were removed

@B2o5T – Because they are old pages that don't exist in English anymore. When we delete (or move) an English page, Crowdin doesn't delete the translated versions from their previous locations. :/

package.json
B2o5T: got it, lgtm

PR #492, opened 2023-08-28
Adding 1 pager

MichaelMacaulay:


website/pages/en/chain-integration-overview.mdx
```diff
@@ -0,0 +1,47 @@
+---
+title: Chain Integration Process Overview
+---
+
+A transparent and governance-based integration process was designed for blockchain teams seeking integration with The Graph protocol. It is a 3-phase process, as summarised below.
+
+## Graph Chain Integration Process
```
benface, line 7: Is this title necessary? Because if it is, the `##`s below should be `###`s, but I would just remove it and then they can remain `##`s.

website/pages/en/chain-integration-overview.mdx
```diff
@@ -0,0 +1,47 @@
...
+
+## Stage 1. Technical Integration
+
+- Teams work on a Graph Node integration. [Here's how](/en/new-chain-integration/).
```
benface, line 11: Links should not include the locale, because otherwise the translated versions of this page will link to an English page.
```suggestion
- Teams work on a Graph Node integration. [Here's how](/new-chain-integration/).
```

website/pages/en/chain-integration-overview.mdx
MichaelMacaulay: No, not 100% necessary. Removing!

website/pages/en/chain-integration-overview.mdx
```diff
@@ -0,0 +1,45 @@
...
+
+---
+
+If the process looks daunting, don't worry! The Graph Foundation is committed to supporting integrators by fostering collaboration, offering essential information, and guiding them through various stages, including navigating governance processes such as Graph Improvement Proposals (GIPs) and pull requests. If you have questions, please reach out to [info@thegraph.foundation](mailto:info@thegraph.foundation) or through Discord (either Pedro, The Graph Foundation member, IndexerDAO, or other core developers).
+
+---
+
+## Frequently Asked Questions
+
+**1. How does this relate to the [World of Data Services GIP](https://forum.thegraph.com/t/gip-0042-a-world-of-data-services/3761)?**
```
benface, line 31: @MichaelMacaulay – I know you merged this already, but I feel like these should be headings?

PR #499, opened 2023-09-01
Remove broken Firehose reference

enoldev:
As a quick fix, let's remove the page since it's no longer available in the Firehose docs.

Next week I'll try to implement an automatic way of tracking changes

### DIFF
--- a/website/pages/en/firehose/_meta.js
+++ b/website/pages/en/firehose/_meta.js
@@ -74,7 +74,7 @@ export default () =>
       items: {
         'design-principles': 'Design Principles',
         'firehose-starter': 'Firehose Acme',
-        'benefits': 'Why Integrate the Firehose',
+        benefits: 'Why Integrate the Firehose',
       },
     },
     references: {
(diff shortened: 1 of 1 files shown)


website/pages/en/firehose/_meta.js
```diff
@@ -74,7 +74,6 @@ export default () =>
       items: {
         'design-principles': 'Design Principles',
         'firehose-starter': 'Firehose Acme',
-        'new-blockchains': 'New Blockchains',
         'why-integrate-the-firehose': 'Why Integrate the Firehose?',
```
benface, line 77:
```suggestion
        'benefits': 'Why Integrate the Firehose',
```
Not sure why this page was renamed because [the `h1` is still "Why Integrate the Firehose"](https://github.com/streamingfast/firehose-docs/blob/master/integrate-new-chains/benefits.md#why-integrate-the-firehose), but it _was_, so.

benface (approved):
Thank you!

PR #517, opened 2023-09-21
Remove Firehose clone

enoldev:


website/pages/en/substreams.mdx
```diff
@@ -2,6 +2,8 @@
 title: Substreams
 ---
 
+![Firehose Logo](/img/substreams-logo.png)
```
benface, line 5:
```suggestion
![Substreams Logo](/img/substreams-logo.png)
```

benface (changes requested):
Looks good, but we also want to update `nginx.conf` to remove the redirect from `/docs/en/firehose/` to `$scheme://$http_host/docs/en/firehose/README/` and perhaps add a redirect from `/docs/en/firehose/(.+)` to `https://firehose.streamingfast.io` like we did with Substreams?

benface (approved):
LGTM, thank you!

PR #526, opened 2023-10-16
Fixing CLI snippets in quick start

MichaelMacaulay:


website/pages/en/quick-start.mdx
```diff
@@ -24,12 +24,16 @@ The Graph CLI is written in JavaScript and you will need to have either `npm` or
 
 On your local machine, run one of the following commands:
 
+Using [npm](https://www.npmjs.com/)
```
benface, line 27: Maybe these should be headings (H3s)? If not, maybe we should add a `:` at the end?

MichaelMacaulay: i think the colons work best, thank you!

benface (approved):

PR #545, opened 2023-11-08
Removing betas & updating based on sunrise

MichaelMacaulay:


website/pages/en/querying/querying-the-hosted-service.mdx
```diff
@@ -19,9 +19,9 @@ This query lists all the counters our mapping has created. Since we only create
 }
 ```
 
-## Using The Hosted Service
+## Using The hosted service
```
benface, line 22: Haha, maybe this one shouldn't be changed, or "The" should be all lowercase too?

website/src/IndexPage.tsx
```diff
@@ -404,8 +404,7 @@ export function SupportedNetworks() {
             )
           })}
         </ul>
-        <Text.P14 color="White64" sx={{ mt: '32px' }}>
-          *{t('index.supportedNetworks.betaWarning')}
+          <Text.P14 color="White64" sx={{ mt: '32px' }}>
         </Text.P14>
```
benface, line 408: Actually can you remove the whole `<Text.P14 color="White64" sx={{ mt: '32px' }}></Text.P14>` (now empty) component please? I would make a suggestion that you can just apply here but it seems I can't due to how the diff looks.

benface (changes requested):
Two small things and good to go.

website/src/IndexPage.tsx
MichaelMacaulay: Sure thign!

benface (approved):
LGTM 👍

PR #557, opened 2023-12-08
Workflow updates and fixes

benface:
This should fix the "Pull Request CI/CD" workflow failing in recent PRs, because it uses the latest LTS version of Node (20) which some of our stuff isn't yet compatible with.


saihaj (approved):

saihaj:
as per this comment, I just upgraded `tsx` safe enough change, if any issue we would see it in build process https://github.com/backstage/backstage/issues/21738#issuecomment-1840776206

PR #562, opened 2023-12-15
fix: table hydration error

saihaj:
https://stackoverflow.com/a/54535317/11321732


benface (approved):
Wow! TIL. Thank you! 🙌

PR #564, opened 2023-12-16
Add `buildGetStaticProps` utility function to be more DRY

benface:
This saves us from having to include `__nextra_pageMap` every time a page needs a `getStaticProps` function, which is error prone because we have to manually pass the locale to `getPageMap()`.

~~Unfortunately, it looks like there's a bug in Nextra that makes the `route` in the `transform` function an empty string for pages with a `getStaticPaths()` function, which is the case of the root ("Get Started") page. As a result, the nav is incorrect for those pages in locales other than English, because it defaults to English. @dimaMachina, if you could take a look it would be very appreciated. 🙏~~  Fixed by @dimaMachina


website/next.config.js
```diff
@@ -20,16 +22,20 @@ const withNextra = nextra({
   codeHighlight: false,
   defaultShowCopyCode: false,
   transform(result, { route }) {
```
dimaMachina, line 24: it's not a bug, because the route is parsed based on filename of md/mdx file

example:
```
/en/my-page.mdx -> /en/my-page
/en/my-folder/index.mdx -> /en/my-folder
/[locale]/index.mdx -> can't parse, route: ''
```

so `buildGetStaticProps` helper function we can use everywhere except dynamic routes aka `[locale]/index.mdx`

benface: Thank you @dimaMachina 🙏

PR #572, opened 2024-01-03
Update new-chain-integration.mdx

pdiomede:
There is a missing word here:

"If you are interested in a different chain type, a new .... with Graph Node must be built"

I suppose it was 'integration', as per below:

"If you are interested in a different chain type, a new integration with Graph Node must be built"


pdiomede:
There is a missing word here:

"If you are interested in a different chain type, a new .... with Graph Node must be built"

I suppose it was 'integration', as per below:

"If you are interested in a different chain type, a new integration with Graph Node must be built"

benface (approved):

benface:
Thank you!

PR #583, opened 2024-01-16
Add docs for indexer hints

incrypto32:


alex-pakalniskis (approved):
content lgtm. thanks @incrypto32

leoyvens:
It would be good to define the meaning of 'history' here and link to the time-travel queries doc.

incrypto32:
@leoyvens 
How does this sound


`history` refers to the historical data maintained by the subgraph, which captures the subgraph's state at each block. This data is essential for time travel queries that require access to the subgraph's previous states at past blocks. For effective use of time travel queries, subgraphs should specify the desired amount of historical data to retain, measured in blocks, or select `never` to keep the entire history.

leoyvens:
I think it's important to add that it specifically refers to old states of mutable entities.

website/pages/en/developing/creating-a-subgraph.mdx
```diff
@@ -150,6 +152,37 @@ The important entries to update for the manifest are:
...
+
+  Using `"auto"` is generally recommended as it maximizes query performance and is sufficient for most users who do not require access to extensive historical data.
+
+  For subgraphs leveraging [time travel queries](/querying/graphql-api/#time-travel-queries), it's advisable to either set a specific number of blocks for historical data retention or use `prune: never` to keep all historical entity states. Below are examples of how to configure both options in your subgraph's settings:
+
+  To retain a specific amount of historical data:
+
+  ```
+   indexerHints:
+    prune: 1000 # Replace 10000 with the desired number of blocks to retain
```
leoyvens, line 176:
```suggestion
    prune: 1000 # Replace 1000 with the desired number of blocks to retain
```

leoyvens (commented):
Nice!

PR #595, opened 2024-01-28
docs: Developer FAQ - Question 10 corrected (Add multiple datasource on graph init command)

shiyasmohd:
On `graph init` command, its now possible to add multiple datasources from `graph-cli` itself. We can also use `graph add` command to add new datasources. This info has not been updated on Question 10 in the FAQ section. 

References - [Link](https://github.com/graphprotocol/graph-tooling/tree/main/packages/cli)


saihaj (approved):
Thank you!

PR #597, opened 2024-02-03
Add 'Handling of entities with identical IDs' section

marcusrein:


leoyvens:
Have you tested this? I think there is a nuance where if the field is explicitly set to `null`, it will be removed in the entity. But if it is simply omited, then it won't affect the enitty.

schmidsi:
I'd add this to "[Writing Mappings](https://thegraph.com/docs/en/developing/creating-a-subgraph/#writing-mappings)" instead of "Defining Entities"

marcusrein:
I tested the `null` handling within a subgraph and updated the description accordingly @leoyvens.

I moved the section to "Writing Mappings" @schmidsi

website/pages/en/developing/creating-a-subgraph.mdx
```diff
@@ -537,6 +533,12 @@ Every entity has to have an `id` that is unique among all entities of the same t
 
 We provide the [Graph Typescript Library](https://github.com/graphprotocol/graph-ts) which contains utilities for interacting with the Graph Node store and conveniences for handling smart contract data and entities. You can use this library in your mappings by importing `@graphprotocol/graph-ts` in `mapping.ts`.
 
+### Handling of entities with identical IDs
+
+When creating and saving a new entity, if an entity with the same ID already exists, the properties of the new entity are preferred during the merge process. This means that the existing entity will be updated with the values from the new entity.
+
+If a null value is set for a field in the new entity with the same ID, the existing entity will be updated with the null value.
+
```
leoyvens, line 541: What if no value is set at all? Good to be explicit about that case.

leoyvens: huh, surprising to me, good that we tested!

leoyvens (approved):

marcusrein:
I added one more paragraph considering this handling @leoyvens, let me know if you approve, thank you.

leoyvens (approved):
lgtm

PR #608, opened 2024-02-21
Update Supported Networks table footnote

benface:
**Before**

CleanShot 2024-02-21 at 18 54 08@2x

**After**

CleanShot 2024-02-21 at 18 54 23@2x


azf20 (approved):
Thanks!

PR #609, opened 2024-02-22
Docs should specify `#` as comment symbol for GraphQL

benbcjdev:
Currently the graphprotocol docs specify double quotations `""` for comments in GraphQL.

According to the GraphQL spec, '#' are used for comments, not double quotes. See the GraphQL spec on comments here:

https://spec.graphql.org/October2021/#sec-Comments


saihaj (approved):
personally I would showcase [block strings for more explanations](https://spec.graphql.org/October2021/#EscapedCharacter) but agreed that `#` should be used for comments as per spec terminology

PR #612, opened 2024-02-26
GraphQL "Int" type is 32-bit (not 3- byte)

benbcjdev:
Per [GraphQL docs](https://graphql.org/learn/schema/#scalar-types), the "Int" type is a signed 32-bit integer. The Graph Node docs currently say it has a size of 32 bytes (might be a typo or something).


saihaj (approved):

saihaj:
Thanks

PR #614, opened 2024-03-01
update "Recommended IDs for New Entities" to "Using Bytes as IDs"

### LINKED ISSUE #613: Update "Recommended IDs for Creating New Entities" to use Bytes as IDs
https://github.com/graphprotocol/docs/blob/main/website/pages/en/developing/creating-a-subgraph.mdx#L530

yash251:
fixes #613.
made use of this [blog](https://thegraph.com/blog/two-simple-subgraph-performance-improvements/)


website/pages/en/developing/creating-a-subgraph.mdx
```diff
@@ -527,13 +527,37 @@ The first handler takes a `NewGravatar` event and creates a new `Gravatar` entit
 
 The second handler tries to load the existing `Gravatar` from the Graph Node store. If it does not exist yet, it is created on-demand. The entity is then updated to match the new event parameters before it is saved back to the store using `gravatar.save()`.
 
-### Recommended IDs for Creating New Entities
+### Using Bytes as IDs
```
schmidsi, line 530: I'd keep the original title. For a new dev, I think this does not make sense. They simply want to learn "The recommended way for IDs"

website/pages/en/developing/creating-a-subgraph.mdx
```diff
@@ -527,13 +527,37 @@ The first handler takes a `NewGravatar` event and creates a new `Gravatar` entit
...
 
-Every entity has to have an `id` that is unique among all entities of the same type. An entity's `id` value is set when the entity is created. Below are some recommended `id` values to consider when creating new entities. NOTE: The value of `id` must be a `string`.
+You can use `Bytes` as the type for the `id` field of entities, and it is highly recommended to use `Bytes` wherever possible, and only use `String` for attributes that truly contain human-readable text, like the name of a token.
```
schmidsi, line 532: I think the alternative to `Bytes` is `ID`, which translates to `String`. Can you validate?

website/pages/en/developing/creating-a-subgraph.mdx
```diff
@@ -527,13 +527,37 @@ The first handler takes a `NewGravatar` event and creates a new `Gravatar` entit
...
 
-- `event.params.id.toHex()`
-- `event.transaction.from.toHex()`
-- `event.transaction.hash.toHex() + "-" + event.logIndex.toString()`
+- Remove unnecessary toHexString() calls. Instead of `transfer.id = event.transaction.hash.toHexString()`,
```
schmidsi, line 534: Again, the mind-set of the reader is that they do not know anything about IDs and want to learn. So there is nothing to "remove"

website/pages/en/developing/creating-a-subgraph.mdx
```diff
@@ -527,13 +527,37 @@ The first handler takes a `NewGravatar` event and creates a new `Gravatar` entit
...
+```typrescript
+let id = event.transaction.hash
+  .toHexString()
+  .concat('-')
+  .concat(BigInt.fromI32(counter).toString())
+```
+
+use `let id = event.transaction.hash.concatI32(counter)`
+
+For entities that store aggregated data, for example, daily trade volumes and the like, the id usually contains the day number. Here, too, using a byte string as the id is beneficial. Determining the id would look like
```
schmidsi, line 547: That's good

website/pages/en/developing/creating-a-subgraph.mdx
```diff
@@ -527,13 +527,37 @@ The first handler takes a `NewGravatar` event and creates a new `Gravatar` entit
...
+
+- Use Bytes for day numbers in aggregated data IDs.
+e.g. 
+```typescript
+let dayID = event.block.timestamp.toI32() / 86400
+let id = Bytes.fromI32(dayID)
+```
+
+- Convert constant addresses to `Bytes`. Here, instead of `const BEEF_ADDRESS = '0xdead...beef'`
+use `const BEEF_ADDRESS = Bytes.fromHexString('0xdead...beef')`
```
schmidsi, line 560: Yes, but we are talking about ids here. I'd do something like:
```
const id = Bytes.fromHexString('0xdead...beef')
```

schmidsi (changes requested):
Great work. I think it's important to have the framing correct: The docs are talking about "now" and not the past and what has changed. Therefor, the best would probably be to assume a new dev reading it. For them, they do not care if there was once a ID that translated to String. They simply want to learn that ids should be 'Bytes` and some common pattern.

website/pages/en/developing/creating-a-subgraph.mdx
yash251: The ID often corresponds to a string value. However, depending on the data source and implementation, it could also be another data type.

schmidsi (changes requested):

website/pages/en/developing/creating-a-subgraph.mdx
```diff
@@ -529,11 +529,22 @@ The second handler tries to load the existing `Gravatar` from the Graph Node sto
...
+```typescript
+let dayID = event.block.timestamp.toI32() / 86400
+let id = Bytes.fromI32(dayID)
+```
+
+- Convert constant addresses to `Bytes`.
+
+`const id = Bytes.fromHexString('0xdead...beef')`
 
 We provide the [Graph Typescript Library](https://github.com/graphprotocol/graph-ts) which contains utilities for interacting with the Graph Node store and conveniences for handling smart contract data and entities. You can use this library in your mappings by importing `@graphprotocol/graph-ts` in `mapping.ts`.
```
schmidsi, line 549: Similarly here but more strict: We can not use "we" language in docs. Who is "we"? The Graph is a protocol. I'd rephrase that paragraph:

> There is the [Graph Typescript Library](https://github.com/graphprotocol/graph-ts) which contains utilities for interacting with the Graph Node store and conveniences for handling smart contract data and entities. It can be imported into `mapping.ts` from `@graphprotocol/graph-ts`.

schmidsi (changes requested):

website/pages/en/developing/creating-a-subgraph.mdx
```diff
@@ -529,11 +529,22 @@ The second handler tries to load the existing `Gravatar` from the Graph Node sto
 
 ### Recommended IDs for Creating New Entities
 
-Every entity has to have an `id` that is unique among all entities of the same type. An entity's `id` value is set when the entity is created. Below are some recommended `id` values to consider when creating new entities. NOTE: The value of `id` must be a `string`.
+You can use `Bytes` as the type for the `id` field of entities, and it is highly recommended to use `Bytes` wherever possible, and only use `String` for attributes that truly contain human-readable text, like the name of a token. Below are some recommended `id` values to consider when creating new entities.
```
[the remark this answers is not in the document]
yash251: noted, thanks for the resource

schmidsi (approved):

PR #617, opened 2024-03-06
Changed invalid 'localhost' URL in grafting.mdx

E-Zbox:
In the "**Building an Existing Subgraph**" section, the URL for 'described more in depth  **_here_**' is invalid.


benface (changes requested):
Thank you! Just left two suggestions; we should use relative URLs instead of absolute since they remain the same across translations.

benface:
Hey, thanks for the PR, but looks like this was fixed in another one because after resolving the conflicts, there are 0 changes here. :P Closing, but thanks again!

PR #625, opened 2024-03-31
docs: Added new Q&A in Developer FAQ

shiyasmohd:
Added new Q&A in Developer FAQ regarding the order of handlers triggered for a data source.


saihaj (approved):

website/pages/en/developing/developer-faqs.mdx
```diff
@@ -136,3 +136,7 @@ The Graph will never charge for the hosted service. The Graph is a decentralized
 ## 27. How do I update a subgraph on mainnet?
 
 If you’re a subgraph developer, you can deploy a new version of your subgraph to the Subgraph Studio using the CLI. It’ll be private at that point, but if you’re happy with it, you can publish to the decentralized Graph Explorer. This will create a new version of your subgraph that Curators can start signaling on.
+
+## 28. In what order are the event, block, and call handlers triggered for a data source?
+
+Event and call handlers are first ordered by transaction index within the block. Event and call handlers within the same transaction are ordered using a convention: event handlers first then call handlers, each type respecting the order they are defined in the manifest. Block handlers are run after event and call handlers, in the order they are defined in the manifest. Also these ordering rules are subject to change.
```
azf20, line 142: @incrypto32 do you think we should add the clarifying note about newly created dynamic data sources? (whose events are processed after the other handlers in a block, which can have confusing side effects)

incrypto32: Yes that's definitely worth mentioning. We could re-word this to something like "Already existing datasources are processed first in the order..." then at the end "then Newly created dynamic data sources are processed at the end, triggers within those datasources follow the same order as above"

Wording needs to be improved, but that's the gist @Shiyasmohd

incrypto32: Or we could do a quote block to specifically mention that newly created dynamic sources are processed last after processing existing ones.

azf20 (approved):
Thanks!

website/pages/en/developing/developer-faqs.mdx
schmidsi: I think this is addressed in https://github.com/graphprotocol/docs/pull/625/commits/0fcfa1127bc651d127d5a17382c487f53a9097b3

PR #631, opened 2024-04-12
Remove querying-with-no-code.mdx and update _meta.js

marcusrein:
Removed Dapplooker from docs.


schmidsi (changes requested):
I'm not sure if this is sufficient. What about this page: https://github.com/graphprotocol/docs/blob/main/website/pages/en/querying/querying-with-no-code.mdx ? @benface, translations are removed simply by removing the `en` page, no? What else is needed?

benface (changes requested):
Looks like you only deleted the page in two languages, not including English.

benface:
> translations are removed simply by removing the en page, no? What else is needed?

@schmidsi – Actually no, we should manually delete pages in all languages. Annoying, and maybe there's a way to change that behavior in Crowdin, but I haven't found it.

benface (approved):

schmidsi (approved):

marcusrein:
This update was performed as Dapplooker is a product that uses The Graph for its service. The Graph docs are meant to be a place for technical information relevant to directly building with The Graph.

PR #643, opened 2024-04-18
Document API versions

azf20:
[Reference](https://github.com/graphprotocol/graph-node/blob/2f55c034348b9ea84f0d3ca2c14bb1918099b253/graph/src/data/subgraph/api_version.rs#L6-L56)

### DIFF
--- a/website/pages/en/developing/graph-ts/api.mdx
+++ b/website/pages/en/developing/graph-ts/api.mdx
@@ -29,6 +29,8 @@ The `apiVersion` in the subgraph manifest specifies the mapping API version whic
 
 | Version | Release notes |
 | :-: | --- |
+| 0.0.9 | Adds new host function `eth_get_balance` |
+| 0.0.8 | Adds validation for existence of fields in the schema when saving an entity. |
 | 0.0.7 | Added `TransactionReceipt` and `Log` classes to the Ethereum types<br />Added `receipt` field to the Ethereum Event object |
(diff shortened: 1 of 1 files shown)


website/pages/en/developing/graph-ts/api.mdx
```diff
@@ -29,6 +29,8 @@ The `apiVersion` in the subgraph manifest specifies the mapping API version whic
 
 | Version | Release notes |
 | :-: | --- |
+| 0.0.9 | Adds new host function `eth_get_balance` |
+| 0.0.8 | Adds validation for existence of fields in the schema when saving an entity. |
```
benface, line 33: Nitpick for consistency with previous entries:
```suggestion
| 0.0.9 | Added new host function `eth_get_balance` |
| 0.0.8 | Added validation for existence of fields in the schema when saving an entity |
```

benface (approved):

website/pages/en/developing/graph-ts/api.mdx
```diff
@@ -29,6 +29,8 @@ The `apiVersion` in the subgraph manifest specifies the mapping API version whic
 
 | Version | Release notes |
 | :-: | --- |
+| 0.0.9 | Adds new host function `eth_get_balance` |
```
Shiyasmohd, line 32: apiVersion 0.0.9 also int

azf20: I think that is the `specVersion` (which is about the structure of the manifest) rather than the `apiVersion`?

PR #654, opened 2024-04-29
Add docs for Topic Filters

incrypto32:


website/pages/en/developing/creating-a-subgraph.mdx
```diff
@@ -785,6 +785,120 @@ indexerHints:
...
+        - event: Approval(address,address,uint256)
+          handler: handleApproval
+        - event: Transfer(address,address,uint256)
+          handler: handleTransfer
+          topic1: ['0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045', '0xc8dA6BF26964aF9D7eEd9e03E53415D37aA96325'] # Optional topic filter which filters only events with the specified topic.
+```
+
+### Indexed Argument Filters / Topic Filters
+
+Topic filters, also known as indexed argument filters, are a powerful feature in subgraphs that allow for precise filtering of blockchain events based on the values of their indexed arguments. These filters are particularly useful for isolating specific events of interest from the vast stream of events on the blockchain, enabling subgraphs to operate more efficiently by focusing only on relevant data. This can be incredibly useful for usecases like creating personal subgraphs that track specific addresses and their interactions with various smart contracts on the blockchain.
```
azf20, line 824: I think worth adding a Solidity snippet to clarify that only `indexed` events can be filtered in this way.
Though does that constraint exist if Graph Node is using a Firehose?

website/pages/en/developing/creating-a-subgraph.mdx
```diff
@@ -785,6 +785,120 @@ indexerHints:
...
+
+#### How Topic Filters Work
+
+When a smart contract emits an event, any arguments that are marked as indexed can be used as filters in a subgraph's manifest. This allows the subgraph to listen selectively for events that match these indexed arguments. The event's first indexed argument corresponds to `topic1`, the second to `topic2`, and so on, up to `topic3`, since the Ethereum Virtual Machine (EVM) allows up to three indexed arguments per event.
```
azf20, line 828: Maybe the snippet could go here. And I think the important thing is that the `indexed` must exist on the original smart contract (developers can't just add it to their ABI)

azf20 (approved):
This is great! Some small suggestions

website/pages/en/developing/creating-a-subgraph.mdx
```diff
@@ -785,6 +785,120 @@ indexerHints:
...
+          file: ./abis/Gravity.json
+      eventHandlers:
+        - event: Approval(address,address,uint256)
+          handler: handleApproval
+        - event: Transfer(address,address,uint256)
+          handler: handleTransfer
+          topic1: ['0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045', '0xc8dA6BF26964aF9D7eEd9e03E53415D37aA96325'] # Optional topic filter which filters only events with the specified topic.
+```
+
+### Indexed Argument Filters / Topic Filters
```
azf20, line 822: Can we add which version this is supported from (specVersion?)

azf20 (changes requested):
This is great but one change request which is to specify the version required to use topic filters

website/pages/en/developing/creating-a-subgraph.mdx
incrypto32: With firehose there is not much benefit of using topic filters. Since firehose doesn't provide a way to filter topics.

azf20 (approved):

PR #661, opened 2024-05-02
remove husky and use CI

saihaj:
Removes husky pre-push commits and just rely on CI jobs for linting/formatting/type checking


package.json
```diff
@@ -24,7 +21,6 @@
     "@edgeandnode/eslint-config": "^2.0.3",
     "eslint": "^8.57.0",
     "eslint-plugin-mdx": "^2.3.4",
-    "husky": "^9.0.11",
     "lint-staged": "^15.2.2",
```
benface, line 24: Should we also remove this?

.github/workflows/ci-cd-pull-request.yml
```diff
@@ -6,9 +6,82 @@ env:
...
+      - name: Install dependencies
+        run: pnpm install
+
+      - name: Typecheck
+        run: pnpm typecheck
+
   build:
     if: contains(github.head_ref, 'crowdin') == false
     runs-on: ubuntu-latest
+    needs: [lint, format, typecheck]
```
benface, line 84: Do we still need to do this step below now that we added these as dependencies?

```
      - name: Lint and typecheck
        run: pnpm check
```

package.json
saihaj: good catch https://github.com/graphprotocol/docs/pull/661/commits/8d2b8ce729c43f9498f720e7e078b60a3e070264

.github/workflows/ci-cd-pull-request.yml
saihaj: yeah not needed

.github/workflows/ci-cd-pull-request.yml
```diff
@@ -6,7 +6,55 @@ env:
...
+        uses: actions/checkout@v4
+
+      - name: Set up env
+        uses: the-guild-org/shared-config/setup@main
+        with:
+          nodeVersion: 20
+          packageManager: pnpm
+
+      - name: Install dependencies
+        run: pnpm install
```
benface, line 52: Are we gonna spend more Actions minutes since this now runs 3 times instead of once?

.github/workflows/ci-cd-pull-request.yml
```diff
@@ -6,7 +6,55 @@ env:
   BASE_IMAGE: ghcr.io/graphprotocol/graph-docs-staging
 
 jobs:
-  build:
+  lint:
```
benface, line 9: This is currently failing on this PR, not sure why. 🤔

benface: @saihaj – No idea why I had to make [these ugly changes](https://github.com/graphprotocol/docs/pull/661/files#diff-056cc9f48e6ecaeee1c02b3bea2b98c1b11453caa18b5f5254f1247cced34c8a) to make `pnpm lint` pass in CI (it works without those changes locally).

benface (approved):

PR #669, opened 2024-05-16
About the graph section changes

idalithb:
Hi! :) 
I re-read this section and eliminated the top description because it seemed quite self-explanatory. I also made minor changes to the sentences. I tried to eliminate complex-compound sentences that slow down the readability.

### DIFF
--- a/website/pages/en/about.mdx
+++ b/website/pages/en/about.mdx
@@ -2,15 +2,14 @@
 title: About The Graph
 ---
 
-This page will explain what The Graph is and how you can get started.
+
 
 ## What is The Graph?
 
 The Graph is a decentralized protocol for indexing and querying blockchain data. The Graph makes it possible to query data that is difficult to query directly.
 
-Projects with complex smart contracts like [Uniswap](https://uniswap.org/) and NFTs initiatives like [Bored Ape Yacht Club](https://boredapeyachtclub.com/) store data on the Ethereum blockchain, making it really difficult to read anything other than basic data directly from the blockchain.
-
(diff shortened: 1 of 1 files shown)


website/pages/en/about.mdx
```diff
@@ -2,33 +2,33 @@
...
+In the case of Bored Ape Yacht Club, we can perform basic read operations on [the contract](https://etherscan.io/address/0xbc4ca0eda7647a8ab7c2061c2e118a18a936f13d#code) like getting the owner of a certain Ape, getting the content URI of an Ape based on their ID, or the total supply. This can be done because these read operations are programmed directly into the smart contract. However, more advanced real-world queries and operations like aggregation, search, relationships, and non-trivial filtering are **not** possible. For example, if we wanted to query for apes that are owned by a certain address and filter by one of its characteristics, we would not be able to get that information by interacting directly with the contract itself.
```
benface, line 13:
```suggestion
In the case of Bored Ape Yacht Club, we can perform basic read operations on [the contract](https://etherscan.io/address/0xbc4ca0eda7647a8ab7c2061c2e118a18a936f13d#code) like getting the owner of a certain Ape, getting the content URI of an Ape based on their ID, or the total supply. This can be done because these read operations are programmed directly into the smart contract. However, more advanced real-world queries and operations like aggregation, search, relationships, and non-trivial filtering are **not** possible. For example, if we wanted to query for Apes that are owned by a certain address and filter by one of its characteristics, we would not be able to get that information by interacting directly with the contract itself.
```

benface (approved):
✨

PR #670, opened 2024-05-16
Getting started edits

idalithb:
Hi! Revised some of the edits and made minor changes to another page.


website/pages/en/about.mdx
```diff
@@ -10,17 +10,17 @@ The Graph is a decentralized protocol for indexing and querying blockchain data.
...
+In the case of Bored Ape Yacht Club, we can perform basic read operations on [the contract](https://etherscan.io/address/0xbc4ca0eda7647a8ab7c2061c2e118a18a936f13d#code) like getting the owner of a certain Ape, getting the content URI of an Ape based on their ID, or the total supply. This can be done because these read operations are programmed directly into the smart contract. However, more advanced real-world queries and operations like aggregation, search, relationships, and non-trivial filtering are *not* possible. For example, if we wanted to query for apes that are owned by a certain address and filter by one of its characteristics, we would not be able to get that information by interacting directly with the contract itself.
```
benface, line 13: Not sure if you didn't see the suggestion I made in #669 or if you just don't agree with it, so going to suggest it again haha:

```suggestion
In the case of Bored Ape Yacht Club, we can perform basic read operations on [the contract](https://etherscan.io/address/0xbc4ca0eda7647a8ab7c2061c2e118a18a936f13d#code) like getting the owner of a certain Ape, getting the content URI of an Ape based on their ID, or the total supply. This can be done because these read operations are programmed directly into the smart contract. However, more advanced real-world queries and operations like aggregation, search, relationships, and non-trivial filtering are *not* possible. For example, if we wanted to query for Apes that are owned by a certain address and filter by one of its characteristics, we would not be able to get that information by interacting directly with the contract itself.
```

benface (approved):

PR #671, opened 2024-05-17
Mitchell/Curation Page Updates

mitchhs12:
Added the support curation grants section, edited and expanded the already existing content, moved bonding curve section  and wrote guide for withdrawing signal.


mitchhs12:
Would appreciate some feedback regarding whether or not I made the right call moving the bonding curve section to the bottom. I think that since the network is on Arbitrum now, it doesn't really apply to the majority of new curators.

azf20 (approved):

PR #674, opened 2024-05-20
add more details on deprecation and rate limiting

azf20:


mitchhs12 (commented):
We've been telling people in the community that the hosted service will be ending and they will have to migrate before 12 June. Most people don't know about the rate limiting and I imagine that they think the hosted service will effectively just stop working on that date.

My only concern is that if we publish more information about this, the urgency to migrate will fade to the back of people's minds and they will just wait until after June 12.

MichaelMacaulay (approved):

PR #675, opened 2024-05-21
add chain overview for indexers

yash251:


website/pages/en/network/indexing.mdx
```diff
@@ -10,6 +10,23 @@ Indexers select subgraphs to index based on the subgraph’s curation signal, wh
 
 <Difficulty level="ADVANCED" />
 
+## Chain Overview
```
azf20, line 13: @Web3Slimchance do you think this is the right page for this in the docs?

PedroMD:
@yash251 @Web3Slimchance , this is great stuff. 

On Starknet, I wonder if it's a good idea to add it to the table, given we'll soon have a better Firehose implementation maintained by the Streaming Fast team (sponsored by the Starknet Foundation itself). 
It may be confusing for Indexers to see it on that page, and I fear many will start getting ready only to see it being deprecated in ~4 weeks (the latest estimate for the new implementation shared by the SF team).
cc @azf20

PedroMD:
Also, do we need the page to go live with the "Additional Info" column, if it's currently completely empty?

Web3Slimchance:
I think the page should be a top-level page under the INDEXING category.

- There isn't a lot of indexer turnover, And existing Indexers are not likely to look deep in the docs to find information about supporting additional networks. (I am afraid that our current indexers won't find it if we put it under "Operating Graph Node). 

- With a rework of the docs on the horizon, I would opt for discoverability over keeping things tidy. 

Open to suggestions for what the name of the page should be.

image

PR #698, opened 2024-06-03
JSON->YAML in glossary.mdx

YaroShkvorets:
Manifest is a YAML file, not JSON.


azf20 (approved):
Thanks! A little known fact is that it actually can be JSON

PR #721, opened 2024-06-28
Update dependencies, including `@edgeandnode/common` + remove `SunriseCountdownBanner`

benface:


website/route-lockfile.txt
saihaj, line 1: maybe we should have a bot to sync whenever we get new translations just to avoid diff changes for other contributors (some people may find it confusing)

saihaj (approved):

website/route-lockfile.txt
benface: Hmm, yeah, there are a few things I want to improve regarding translations. Another one is that we can't translate titles defined in `_meta.js`. If I'm not mistaken, @dimaMachina told me that would get fixed when we migrate to Nextra 3. Maybe we can also look into it then? Another one, potentially quite serious, is that anchor links are broken in translated pages, because the heading IDs are generated from the translated text, instead of the English text. I'm really not sure how to fix that one. :/

PR #732, opened 2024-07-29
2/4 move `@graphprotocol/nextra-theme` to `@graphprotocol/docs`

dimaMachina:


saihaj (approved):

benface (approved):
Sweet!

Just one thing, and maybe it's just me, but should `pages` go inside `src`? Seems a bit strange to me to have them as siblings.

dimaMachina:
> Sweet!
> 
> Just one thing, and maybe it's just me, but should `pages` go inside `src`? Seems a bit strange to me to have them as siblings.

yes, better to put them under `src`

dimaMachina:
@benface done here https://github.com/graphprotocol/docs/pull/734

PR #504, opened 2023-09-06
1/4 Nextra v3 with i18n by locale folders

dimaMachina:
This PR includes migration to Nextra 3 with builtin support i18n by folders. More info about changes from v2 can be found in blogpost https://the-guild.dev/blog/nextra-3

also there is support of changing sidebar title via frontmatter prop

```md
---
title: My page title
sidebarTitle: My sidebar title, otherwise fallback to frontmatter.title
---
```

PR #733, opened 2024-07-30
3/4 Refactor `_meta.js` to prefer `sidebarTitle`

dimaMachina:


website/next.config.js
```diff
@@ -23,56 +21,85 @@ const env = {
...
+      developing: t('developing'),
+      deploying: t('deploying'),
+      publishing: t('publishing'),
+      managing: t('managing'),
+      querying: t('querying'),
+      cookbook: t('cookbook'),
+      'release-notes': t('releaseNotes'),
+      '---3': {
+        type: 'separator',
+      },
```
dimaMachina, line 69: @benface now root _meta file will be translated by Crowdin, I remember you asked it in the past

website/next.config.js
```diff
@@ -23,56 +21,85 @@ const env = {
...
-            if (node.url.startsWith('../')) {
-              const GO_BACK_REPEATED_REGEX = /(\.\.\/)+/
-              node.url = node.url.replace(GO_BACK_REPEATED_REGEX, `https://github.com/${user}/${repo}/tree/${branch}/`)
-            }
-          })
-          remarkReplaceLinks({ foundPath: fileName, basePath })(tree, file)
-        }
-      },
-    ],
+    remarkPlugins: [remarkTransformRemoteGitHub],
```
dimaMachina, line 102: moved all code directly in remark plugin to simplify next.config file

website/src/remarkTransformRemoteGitHub.js
```diff
@@ -0,0 +1,67 @@
...
+      // Relative to the current path, e.g. (foo)[bar] or (foo)[./bar] or (foo)[../bar]
+      node.url = path.join(basePath + path.dirname(fileName), node.url)
+    }
+  })
+
+  visit(tree, { type: 'heading', depth: 1 }, (node) => {
+    // Replace h1 with YAML frontmatter
+    node.type = 'yaml'
+    node.value = `title: "${node.children[0].value}"
+remoteFilePath: https://github.com/${user}/${repo}/tree/${branch}/${docsPath}${fileName}`
```
dimaMachina, line 60: edit on github will point to right github file

website/next.config.js
benface: This is awesome @dimaMachina, thank you!

website/pages/_app.tsx
```diff
@@ -17,11 +17,11 @@ import {
 import { CookieBanner, GlobalFooter, GlobalHeader } from '@edgeandnode/go'
 
 import { supportedLocales, translations, useI18n } from '@/i18n'
-import { DocSearch } from '@/src/nextra-theme/components'
+import { DocSearch } from '@/src/components'
 
 import '@edgeandnode/gds/style.css'
 import '@docsearch/css'
-import '../src/app.css'
+import './globals.css'
```
benface, line 24: Just a preference thing?

website/pages/en/global.json
```diff
@@ -10,5 +10,20 @@
   "notFoundTitle": "Oops! This page was lost in space...",
   "notFoundSubtitle": "Check if you’re using the right address or explore our website by clicking on the link below.",
   "goHome": "Go Home",
-  "video": "Video"
+  "video": "Video",
+  "sidebar": {
```
benface, line 14: Should this be named `nav` instead of `sidebar`?

benface: Hmm, I guess `sidebar` is consistent with Nextra's naming of `sidebarTitle`. Just seems a little too concerned about the UI. :P

website/next.config.js
```diff
@@ -23,56 +21,85 @@ const env = {
...
-    const locale = pageMap[0].data.slice(0, 2)
+    const locale = pageMap[0].route.slice(1, 3)
+    // TODO: remove this when crowdin will translate global.json
+    const t = (key) => {
+      try {
+        return translate(translations, locale, `global.sidebar.${key}`)
+      } catch {
+        return translate(translations, 'en', `global.sidebar.${key}`)
+      }
+    }
```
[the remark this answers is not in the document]
dimaMachina: for now, some keys are missing translation and I have a runtime error

website/pages/_app.tsx
dimaMachina: yeah, similar to structure like in `app` router, can do rollback of this change

website/next.config.js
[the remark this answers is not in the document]
benface: Weird, `translate` should fall back to English if a key is missing in the translation.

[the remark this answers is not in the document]
dimaMachina: I had error like value is not string

website/pages/_app.tsx
benface: Ah, no, if that's how Next calls it by default, let's go with that. (By the way, does Nextra support the `app` router? If so, do you think we should switch to it?)

website/next.config.js
```diff
@@ -23,56 +21,85 @@ const env = {
...
     return [
-      ...pageMap,
+      { data: metaFile },
       {
         route: `/${locale}`,
         name: 'index',
-        frontMatter: {
-          title: translate(translations, locale, 'index.title'),
-        },
+        frontMatter: {},
```
benface, line 96: I'm even more confused about this now because didn't you say [here](https://github.com/graphprotocol/docs/pull/504#discussion_r1696702807) that this page map entry was needed to translate the title, and now you remove the translation?

website/pages/_app.tsx
dimaMachina: Nextra v4 will support only `app` router + server side i18n. I did like 80% of the migration in January. After worked on GraphQL.org migration to Nextra and new features to Hive

website/next.config.js
dimaMachina: I removed it here because I added in `meta` object https://github.com/graphprotocol/docs/blob/f5f1cb0382894a41c5436928b0e3b049bfc9320e/website/next.config.js#L40-L41

there is no need to duplication with this change

benface: Ah got it. But what happens if this page map entry is removed? The index page will 404?

dimaMachina: I guess it is to pass this check https://github.com/graphprotocol/docs/blob/5634476928786ea9aba552dd17c19574885ebee8/packages/nextra-theme/src/index.tsx#L104-L111

website/next.config.js
```diff
@@ -23,56 +21,85 @@ const env = {
...
 const withNextra = nextra({
-  theme: './src/nextra-theme/index.tsx',
+  theme: './src/NextraLayout.tsx',
   search: false,
   codeHighlight: false,
   defaultShowCopyCode: false,
   readingTime: true,
   transformPageMap(pageMap) {
-    const locale = pageMap[0].data.slice(0, 2)
+    const locale = pageMap[0].route.slice(1, 3)
```
benface, line 30: Hey @dimaMachina I'm getting this error after checking out this branch and running `pnpm install` and `pnpm build`:

> TypeError: Cannot read properties of undefined (reading 'route')

If I `console.log({ pageMap })` before this line, I get `{ pageMap: [] }`. What am I doing wrong?

PR #734, opened 2024-07-30
4/4 move `pages` to `src/pages`

dimaMachina:
to simplify review, just review last commit 😆


website/src/contracts.tsx
```diff
@@ -17,19 +17,17 @@ export function ProtocolContractsTable({ networkId }: { networkId: number }) {
     <Table>
       <tbody>
         <tr>
-          <th>{'Contract'}</th>
-          <th>{'Address'}</th>
+          <th>Contract</th>
+          <th>Address</th>
```
benface, line 21: These should probably be translated, adding myself a TODO for that.

benface (approved):
Thank you! 🙇‍♂️

website/src/contracts.tsx
dimaMachina: want to rollback?

benface: No no, as it is now, it makes no sense to use interpolation. This is a good change. :P

PR #736, opened 2024-08-07
Update quick-start.mdx

pfmescher:
Added link to graphiql playground

### DIFF
--- a/website/pages/en/quick-start.mdx
+++ b/website/pages/en/quick-start.mdx
@@ -98,7 +98,7 @@ The logs will tell you if there are any errors with your subgraph. The logs of a
 
 ![Subgraph logs](/img/subgraph-logs-image.png)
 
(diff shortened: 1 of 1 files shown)


saihaj (changes requested):
Best way right now would be checking the graph explorer. The endpoint being proposed is from Hosted service which is now sunsetted. Even if we can return status that would only be Studio subgraph and not everything on the network.


@idalithb we should re-visit this section. [Internal discussion](https://guild-oss.slack.com/archives/C020JKTV7L4/p1723061885397109?thread_ts=1723052350.959129&cid=C020JKTV7L4) for context

0xa3k5:
Thank you @pfmescher  for your contribution and for taking the time to improve the docs. Since this PR was opened, the docs has undergone a major redesign with significant structural and content changes. Because of that, this PR is now out of sync with the current codebase and can't be merged as-is. I'm closing this PR.

PR #740, opened 2024-08-13
Scrub HS from docs

MichaelMacaulay:

PR #749, opened 2024-08-22
fix: remove dead sidebar links

DenhamPreen:
Looks like this might have been missed here
https://github.com/graphprotocol/docs/pull/740/files cc: @MichaelMacaulay

These links are dead
image


benface:
Looks like this was already handled in the PR just before this one, #748, so I'm going to close this. But thank you very much!

PR #750, opened 2024-08-22
Fixing typos

MichaelMacaulay:


benface (approved):
Wow, good catches. Thanks for doing this.

PR #741, opened 2024-08-13
Managing subgraphs

idalithb:
-combined two pages

PR #753, opened 2024-08-22
Clean up redirects a little bit

benface:
This includes the missing redirects from #741 as well as some other redirect-related changes (will comment inline).


nginx.conf
```diff
@@ -44,10 +44,10 @@ http {
         rewrite ^/(.*)//+(.*)   $scheme://$http_host/$1/$2  permanent; # remove adjacent slashes in the middle of the URI
         rewrite ^/([^.]*[^/])$  $scheme://$http_host/$1/    permanent; # force a trailing slash except for files with an extension
 
-        # Permanent redirects (301)
-        rewrite ^/docs/hostedservice/(.*)$                              $scheme://$http_host/docs/en/hosted-service/$1      permanent;
```
benface, line 48: Those `/docs/hostedservice/*` URLs must be really old; I couldn't find any reference to those anywhere. Also, we deprecated the HS, so I think it's safe to remove this rule.

nginx.conf
```diff
@@ -88,14 +88,16 @@ http {
         rewrite ^/docs/([a-zA-Z][a-zA-Z])/arbitrum-faq/$                                $scheme://$http_host/docs/$1/arbitrum/arbitrum-faq/                         permanent;
         rewrite ^/docs/([a-zA-Z][a-zA-Z])/cookbook/migrating-a-subgraph/$               $scheme://$http_host/docs/$1/cookbook/upgrading-a-subgraph/                 permanent;
         rewrite ^/docs/([a-zA-Z][a-zA-Z])/cookbook/quick-start/$                        $scheme://$http_host/docs/$1/quick-start/                                   permanent;
-        rewrite ^/docs/en/substreams/(?!index\.).+$                                     https://substreams.streamingfast.io                                         permanent;
-        rewrite ^/docs/en/firehose/(?!index\.).+$                                       https://firehose.streamingfast.io                                           permanent;
```
benface, line 92: Just moved these to the bottom of the list because they're different from the other (English-only).

nginx.conf
```diff
@@ -88,14 +88,16 @@ http {
...
         rewrite ^/docs/([a-zA-Z][a-zA-Z])/developer/assemblyscript-api/$                $scheme://$http_host/docs/$1/developing/graph-ts/api/                       permanent;
         rewrite ^/docs/([a-zA-Z][a-zA-Z])/developing/assemblyscript-api/$               $scheme://$http_host/docs/$1/developing/graph-ts/api/                       permanent;
+        rewrite ^/docs/([a-zA-Z][a-zA-Z])/managing/deprecating-a-subgraph/$             $scheme://$http_host/docs/$1/managing/transfer-and-deprecate-a-subgraph/    permanent;
+        rewrite ^/docs/([a-zA-Z][a-zA-Z])/managing/transferring-subgraph-ownership/$    $scheme://$http_host/docs/$1/managing/transfer-and-deprecate-a-subgraph/    permanent;
```
benface, line 94: These are the redirects that should have been included in #741; my bad for not noticing in my review. :)

website/next.config.js
```diff
@@ -56,31 +56,7 @@ export default withNextra({
...
-    {
-      source: '/en/developer/assemblyscript-api/',
-      destination: '/en/developing/graph-ts/api/',
-      permanent: true,
-    },
-    {
-      source: '/en/developing/assemblyscript-api/',
-      destination: '/en/developing/graph-ts/api/',
-      permanent: true,
-    },
```
benface, line 83: The only reason we ever needed to include redirects here (in addition to `nginx.conf)` is that there can still be links to the old URLs in the Markdown files. If that is the case, Next needs to be aware of them, otherwise the client-side router will link to a 404 (or maybe attempt to do a full page reload, which would be better, but still not ideal). But I hate having to maintain the same list of redirects in two places, so I went ahead and [replaced](https://github.com/graphprotocol/docs/pull/753/commits/a745368b20f96ab27eb339b46d68d76e2e13136a) all the old URLs still lingering around to the new ones, which allows us to remove these.

cmwhited (approved):

PR #764, opened 2024-09-11
removal

idalithb:
-Removing a straggler


benface:
I think you accidentally deleted the `vi` version of the page instead of the `en` one :P

idalithb:
I was attempting to delete this page: https://thegraph.com/docs/vi/deploying/hosted-service/

benface:
Oh, I guess I don't understand why that one in particular? I see now that the page was already deleted in English, but not in any of the other languages.

idalithb:
Oh, I understand now. Will it automatically be deleted in all other languages?

PR #774, opened 2024-09-18
Redirect TAP Version to indexer docs

gusinacio:
Updated the `en/tap.mdx` file and copied it to all other languages.


MichaelMacaulay (approved):
LGTM

PR #791, opened 2024-10-08
Update dependencies to refresh Supported Networks (+ sort them by name)

benface:
CleanShot 2024-10-08 at 12 34 29@2x


idalithb (approved):
TY!!! @benface

PR #794, opened 2024-10-14
Update

MichaelMacaulay:


benface (approved):
We'll have to remember to update this when it does become supported in Graph Node.

PR #802, opened 2024-10-22
Update dependencies to refresh supported networks (including logos)

benface:
CleanShot 2024-10-22 at 19 17 22@2x

### DIFF
--- a/website/src/IndexPage.tsx
+++ b/website/src/IndexPage.tsx
@@ -203,7 +203,7 @@ export function SupportedNetworks() {
         .map((network) => (
           <Flex.Column key={network.uid} as="li" align="center" gap={Spacing['8px']} sx={{ color: 'White64' }}>
             <div sx={{ height: '40px' }}>
-              <NetworkIcon chain={network as any} size="40px" title="" />
+              <NetworkIcon chain={network} size="40px" />
             </div>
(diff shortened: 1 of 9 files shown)


idalithb (approved):

website/src/IndexPage.tsx
```diff
@@ -203,7 +203,7 @@ export function SupportedNetworks() {
         .map((network) => (
           <Flex.Column key={network.uid} as="li" align="center" gap={Spacing['8px']} sx={{ color: 'White64' }}>
             <div sx={{ height: '40px' }}>
-              <NetworkIcon chain={network as any} size="40px" title="" />
```
cmwhited, line 206: We do have some utils in common to validate that a string is a valid NetworkChainUID value that could use here as well

cmwhited (approved):

website/src/IndexPage.tsx
benface: We don't really need to validate here; `NetworkIcon` will fail silently / return `null` if for some reason `network` is not a valid chain UID (but it really should be since [it comes from `@edgeandnode/common`](https://github.com/graphprotocol/docs/blob/main/website/src/supportedNetworks.tsx#L28)).

PR #820, opened 2024-11-19
Update indexing.mdx

denverbaumgartner:
correct number value referring to allocation states


MichaelMacaulay (approved):
Thank you for catching! LGTM

PR #822, opened 2024-11-22
Managing API Keys-Update 

idalithb:
- Updated copy 
- Let me know if I anything is inaccurate, can be removed from this page and placed under billing, or if it can simply be deleted. TY 😄

### DIFF
--- a/website/pages/en/querying/managing-api-keys.mdx
+++ b/website/pages/en/querying/managing-api-keys.mdx
@@ -29,7 +29,6 @@ You can click on an individual API key to view the Details page:
    - View the current usage of the API key with stats:
      - Number of queries
      - Amount of GRT spent
-2. Under the **Security** section, you can:
-   - Opt into security settings depending on the level of control you’d like to have
+2. Under the **Security** section, you can opt into security settings depending on the level of control you’d like to have. Specifically, you can:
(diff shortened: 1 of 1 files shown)


website/pages/en/querying/managing-api-keys.mdx
```diff
@@ -2,23 +2,34 @@
...
-1. The **Overview** section will allow you to:
+1. Under the **Overview** section, you can:
    - Edit your key name
    - Regenerate API keys
    - View the current usage of the API key with stats:
      - Number of queries
      - Amount of GRT spent
-2. Under **Security**, you’ll be able to opt into security settings depending on the level of control you’d like to have over your API keys. In this section, you can:
+2. Under the **Security** section, you can:
+   - Opt into security settings depending on the level of control you’d like to have
```
benface, line 33: One of these bullet points is not like the others, as in, this is more of a summary of the following ones instead of a feature in itself. So I would keep it as the introduction, but with your better phrasing:
```suggestion
2. Under the **Security** section, you can opt into security settings depending on the level of control you’d like to have. Specifically, you can:
```

idalithb: I like it!

benface (approved):

PR #833, opened 2024-12-10
Adding Deprecated Message

idalithb:
- Note at the top of each page before they all get archived


benface (approved):

0xa3k5:
is this still relevant @idalithb ? If so, I can add it to #946 otherwise we can close this PR.

PR #838, opened 2024-12-14
Update dependencies + add `fix-pages-structure` script and run it before build

benface:
This PR applies minor dependency updates, which notably refreshes the Supported Networks table, and it adds a `fix-pages-structure` script to improve the process of adding/removing/moving pages, considering our Crowdin setup. It ensures that any page or directory that was added or removed in English is also added or removed in other languages. The script runs on `prebuild` so it will run in CI/CD, preventing build errors in PRs that add or move pages, which has been a recurring source of frustration. I also ran it and committed the results directly in this PR; you can see that it removes a bunch of orphaned pages and adds some missing `_meta.js` files.


idalithb:
@benface Terrific work! This is such an improvement 🥳

PR #854, opened 2025-01-16
style updates

0xa3k5:
- bumped font sizes (navbar + prose)
- bumped icon sizes (navbar)
- set max width for article

### DIFF
--- a/website/src/NextraLayout.tsx
+++ b/website/src/NextraLayout.tsx
@@ -629,7 +629,8 @@ export default function NextraLayout({ children, pageOpts, pageProps }: NextraTh
               <article
                 className={`
                   text-p20 mx-auto max-w-[min(50vw,80ch)]
-                  leading-9 text-white/64 mdx-[h1]:text-h40
+                  leading-9 text-white/64 
+                  mdx-[h1]:text-h40
                   mdx-[h2]:text-h32
                   mdx-[h3]:text-h24
                   mdx-[h4,h5,h6]:text-h20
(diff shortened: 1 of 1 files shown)


website/src/NextraLayout.tsx
```diff
@@ -625,14 +625,15 @@ export default function NextraLayout({ children, pageOpts, pageProps }: NextraTh
...
-                  mdx-[h1]:text-h24
-                  mdx-[h2]:text-h20
-                  mdx-[h3]:text-h18
-                  mdx-[h4,h5,h6]:text-h16
+                  text-p20 mx-auto max-w-[min(50vw,80ch)]
+                  leading-9 text-white/64 
+                  mdx-[h1]:text-h40
+                  mdx-[h2]:text-h32
+                  mdx-[h3]:text-h24
+                  mdx-[h4,h5,h6]:text-h20
```
benface, line 636: Should `h5` and `h6` be 18px? instead of 20px?

website/src/NextraLayout.tsx
```diff
@@ -625,14 +625,15 @@ export default function NextraLayout({ children, pageOpts, pageProps }: NextraTh
 
           <main>
             <header className="h-16 border-b border-white/8"></header>
-            <div className="px-26 py-12">
+            <div className="w-full px-8 py-12 md:px-26">
```
benface, line 628: This `w-full` doesn't seem necessary.

benface (approved):
🙏

website/src/NextraLayout.tsx
0xa3k5: I think 20 is good, but let's keep iterating and testing it

benface: I guess that means we should recommend not using `h5` or `h6` because they don't look any different than `h4` so it could be confusing, right? 3 levels of headings + the title seems reasonable to me.

PR #871, opened 2025-02-13
fix `editPageUrl`

0xa3k5:


0xa3k5:
closed because this is going to be solved when the v2 branch is merged, only a problem in the staging environment

PR #872, opened 2025-02-13
style tweaks for home page

0xa3k5:
- left align the hero
- remove gradient from the hero title
- update the background svg to accommodate these changes
- reduce the height of the line for the divider between "g" and "docs" on logo
- increase the height of the divider after the sidebar toggle button to full span
- add icons to all products in the graph products section


benface (approved):
Awesome, thank you!

PR #873, opened 2025-02-13
Redesign (v2)

benface:
https://github.com/user-attachments/assets/ca44c879-6f59-4279-b031-fd03663f34b5


0xa3k5 (approved):

website/tailwind.config.mjs
```diff
@@ -33,31 +33,31 @@ export default buildTailwindConfig({
...
-          '@apply font-regular text-pretty text-18 leading-8': '',
+          '@apply text-pretty font-regular text-18 leading-8': '',
         },
         '.text-body-small': {
-          '@apply font-regular text-pretty text-16': '',
+          '@apply text-pretty font-regular text-16': '',
         },
         '.text-body-xsmall': {
-          '@apply font-regular text-pretty text-14': '',
+          '@apply text-pretty font-regular text-14': '',
```
0xa3k5, line 60: curious, why are we not using text-pretty on the root or on the Layout?

benface: At this point, we could, and I would if making another commit didn't force me to wait another 20 minutes for CI to complete haha. But the reason why I didn't do it was that in principle, we want to be intentional about where to use `text-wrap: pretty` and `text-wrap: balance` because they have performance implications, but in reality, those text styles are used in 90%+ of the elements on the page, so it would really not make a difference. I will clean it up though, promise.

PR #875, opened 2025-02-14
tweak folder structure

0xa3k5:
- create a new "Best Practices" folder under the Subgraphs and move the best practice pages to that folder. (these pages were under subgraphs > cookbooks)
- rename the Best Practice pages' sidebar title to not include the "Best Practice 1/2/3/4/5/6 in them. (Best Practices 1: Pruning with indexerHints -> Pruning with indexerHints etc)
- slightly widen the sidebar width 
- rename "Release Notes and Upgrade Guides" to Migration Guides (there is only 2 pages under that folder and both of them mentions migration guides instead of upgrade guides)
- move Subgraph Studio FAQ to Resources section (was under the Subgraphs > Developing > Deploying (I thought its better under resources because it was previously very buried and looks like its more general FAQ than just deploying)
- Slightly shorten a few sidebar titles (while keeping the titles meaningful and respected to how it was before)


benface:
Thanks so much AK, and sorry I didn't figure out how to keep your commits. 😅 Superseded by #885

PR #879, opened 2025-02-15
Comp-Updated

idalithb:
- Updating previous PR draft to avoid broken links and conflicts. 
- Please review **EN version** of this page _only_


website/src/pages/ar/subgraphs/cookbook/subgraph-composition.mdx
benface, line 1: Hey @idalithb, I notice this PR adds the same page in every language. I'm guessing you ran the `fix-pages-structure` script? I would recommend not doing that until right before merging the PR (so after it's been approved), otherwise it makes it a bit harder to review and even harder for you to address changes from the review (you can of course only address the changes in English and the other pages will be updated whenever we merge the next Crowdin PR, but it does create potential confusion in the meantime).

website/src/pages/en/subgraphs/cookbook/subgraph-composition.mdx
```diff
@@ -0,0 +1,132 @@
+---
+title: Subgraph Composition with Sushiswap v3 on Base
+sidebarTitle: 'Subgraph Composition'
+---
+
+Leverage subgraph composition to speed up development time. Create a base subgraph with essential data, then build additional subgraphs on top of it.
```
benface, line 6:
```suggestion
Leverage Subgraph composition to speed up development time. Create a base subgraph with essential data, then build additional subgraphs on top of it.
```
I think we should always capitalize Subgraph now (since #886 was merged). I won't leave the same comment for every mention of "subgraph" below, but it should be applied to the whole document. cc @MichaelMacaulay

website/src/pages/en/subgraphs/cookbook/subgraph-composition.mdx
```diff
@@ -0,0 +1,132 @@
...
+
+While the source subgraph is a standard subgraph, the dependent subgraph utilizes the subgraph composition feature.
+
+### Source Subgraph
+
+The source subgraph tracks events from the Sushiswap v3 subgraph on the Base chain. This subgraph's configuration file is in the `source/subgraph.yaml`.
+
+### Dependent Subgraph
+
+The dependent subgraph is in the `dependent/subgraph.yaml` directory, which specifies the source subgraph as a data source. This subgraph uses entities from the source to trigger specific actions based on changes to those entities.
```
benface, line 48: `dependent/subgraph.yaml` is a file, not a directory.

website/src/pages/en/subgraphs/cookbook/subgraph-composition.mdx
```diff
@@ -0,0 +1,132 @@
...
+
+## Get Started
+
+The following is a how-to-guide on using one subgraph as a data source for another.
```
benface, line 52:
```suggestion
The following is a how-to guide on using one Subgraph as a data source for another.
```

website/src/pages/en/subgraphs/cookbook/subgraph-composition.mdx
```diff
@@ -0,0 +1,132 @@
...
+
+- Use this powerful tool to scale your subgraph development and reuse existing data.
+- The setup includes creating a base source subgraph and referencing it in a dependent subgraph.
+- You define handlers in the dependent subgraph to perform actions based on changes in the source subgraph's entities.
+
+This approach unlocks composability and scalability, simplifying both development and maintenance efficiency.
+
+## Additional Resources
+
+To use other advance features in your subgraph check out [subgraph advance features](/developing/creating/advanced/) and the [subgraph composition repo](https://github.com/incrypto32/subgraph-composition-sample-subgraph).
```
benface, line 130:
```suggestion
To use other advanced features in your Subgraph, check out [Subgraph advanced features](/developing/creating/advanced/) and the [Subgraph composition example repo](https://github.com/incrypto32/subgraph-composition-sample-subgraph).
```

website/src/pages/en/subgraphs/cookbook/subgraph-composition.mdx
```diff
@@ -0,0 +1,132 @@
...
+
+To learn how to define three `source` subgraphs check out the [subgraph composition example repo](https://github.com/isum/subgraph-composition-example).
```
benface, line 132:
```suggestion
To learn how to define three `source` Subgraphs, check out [this Subgraph composition example repo](https://github.com/isum/subgraph-composition-example).
```

website/src/pages/ar/subgraphs/cookbook/subgraph-composition.mdx
benface: By the way, you don't technically need to run `fix-pages-structure` at all when adding a new page (notice I'm not mentioning it in [the "Adding new pages" section of the CONTRIBUTING guide](https://github.com/graphprotocol/docs/blob/main/CONTRIBUTING.md#adding-new-pages)), because it is run automatically during the build in CI/CD. It is appreciated, because those files do need to be checked in git when they get translated and it's better if that's done by the PR that introduces the page, but right before merging is definitely ideal for the reasons mentioned above. So in practice, it's probably simpler if you don't run it at all and those files get added by the Crowdin PR. Sorry, I know it's complicated!

idalithb: @benface Ah! That makes sense. Yes, this was before I read your updates to that section. Will it create additional conflicts now with this PR?

website/src/pages/en/subgraphs/cookbook/subgraph-composition.mdx
```diff
@@ -0,0 +1,132 @@
+---
+title: Subgraph Composition with Sushiswap v3 on Base
+sidebarTitle: 'Subgraph Composition'
+---
+
+Leverage subgraph composition to speed up development time. Create a base subgraph with essential data, then build additional subgraphs on top of it.
+
+> - Subgraph composition is built into the CLI, and you can deploy with [Subgraph Studio](https://thegraph.com/studio/).
+> - You can use existing subgraphs, but they must be redeployed with a new spec version, which doesn't require you to write new code.
+> - You may want to restructure your subgraph to split out the logic as you move to a composable subgraph world.
```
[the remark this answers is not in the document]
idalithb: @benface I like how it looks. Will look it over and implement markdown--TY!

website/src/pages/en/subgraphs/cookbook/subgraph-composition.mdx
idalithb: YES!

website/src/pages/en/subgraphs/cookbook/subgraph-composition.mdx
idalithb: I believe this was a suggested edit by @MichaelMacaulay, lmk if you're good to change back to file.

benface: Or alternatively:
```suggestion
The dependent subgraph is in the `dependent` directory, which specifies the source subgraph as a data source. This subgraph uses entities from the source to trigger specific actions based on changes to those entities.
```

website/src/pages/ar/subgraphs/cookbook/subgraph-composition.mdx
benface: No, the problem is not conflicts, it's just that if you make changes to address e.g. my feedback, you have to make sure to do them in English and not to one of the other files that are identical, and then either delete all those other files and re-run the script so they all have the updated content, or (more realistically) leave them be (it's really fine, as they will be overwritten with either the translated content or an exact copy of the English content when the Crowdin PR is merged) but then it becomes hard/confusing for future reviewers of this PR to see the latest content because English is not the first one that appears when reviewing the changed files.

benface: I would say don't worry about it for now, but yeah, my latest recommendation is just to never run that script. Now that it's run in CI/CD, the build should never break by adding a new file only in English. :)

MichaelMacaulay (approved):
LGTM

PR #882, opened 2025-02-18
remove cosmos

yash251:


yash251:
cc @benface @idalithb

benface (approved):
🙌

benface:
Oh @yash251, I see there's a formatting issue which makes the `format` CI check fail. We use Prettier; I recommend installing the VS Code Prettier extension and enabling "Format on save" to not have to think about it. But I will run `pnpm check:fix` in your branch, it should fix it. Thank you!

PR #890, opened 2025-02-27
Killing cookbook

MichaelMacaulay:


website/src/pages/en/subgraphs/_meta.js
```diff
@@ -6,6 +6,6 @@ export default {
   querying: titles.querying,
   developing: titles.developing,
   billing: '',
-  cookbook: titles.cookbook,
+  guides: 'How-to Guides',
```
benface, line 9: Folder titles should really live in the `_meta-titles.json` files to be translatable. 🙏

benface: Ah, looks like you already added it there. So:
```suggestion
  guides: titles.guides,
```
You may need to add it to all the languages, though, otherwise this will be `undefined` for some of them and I think that will break the build (maybe that's why you hardcoded it as a string here?)

benface: Oh, you added it to the wrong `_meta-titles.json`. 😅 It should be in the one in the `subgraphs` folder, and there should already be a `cookbook: 'Cookbook'` entry there for you to change.

website/src/pages/en/subgraphs/developing/_meta-titles.json
```diff
@@ -2,5 +2,6 @@
   "creating": "Creating",
   "deploying": "Deploying",
   "publishing": "Publishing",
-  "managing": "Managing"
+  "managing": "Managing",
+  "guides": "How-to Guides"
```
benface, line 6: Like I wrote above, looks like you modified the wrong `_meta-titles.json`.

benface (changes requested):
Left some comments, but also, we should add redirects for all the pages under `/subgraphs/cookbook` that got moved to `/subgraphs/guides`.

website/src/pages/en/subgraphs/cookbook/subgraph-uncrashable.mdx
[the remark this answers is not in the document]
idalithb: yes! @benface Did we add an explanation of redirects in the contribution notes?

idalithb (commented):
- After you address @benface's comments merge this baby! I'm so excited to have this change 👏

website/src/pages/en/subgraphs/cookbook/subgraph-uncrashable.mdx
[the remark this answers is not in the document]
benface: There's a mention of it [here](https://github.com/graphprotocol/docs/blob/main/CONTRIBUTING.md#moving-or-renaming-pages-or-directories), but I don't know if I would call it an "explanation" haha. I didn't want to go too much in the technical weeds in the CONTRIBUTING guide since writing Nginx rewrite rules is a whole topic, but maybe we could link to [these docs](https://nginx.org/en/docs/http/ngx_http_rewrite_module.html)? It's also very much possible to create a redirect without understanding any of it by just copying & pasting another one and modifying it slightly, potentially with the help of AI.

PR #891, opened 2025-02-28
reading time & last updated indicator designs

0xa3k5:
- add Hourglass icon for reading time
- add Calendar icon for last updated
- make "edit this page" always on the bottom of the page (remove it from the TOC)

screenshot@2x
screenshot@2x
screenshot@2x
screenshot@2x
screenshot@2x
screenshot@2x


website/src/Layout.tsx
```diff
@@ -710,32 +712,18 @@ function MDXContent({ toc: headings, children }: ComponentPropsWithoutRef<Nextra
           ) : null}
           {!frontMatter.hideContentHeader ? (
             <header className="col-[container] pt-12 transition-[padding] duration-[var(--graph-docs-layout-transition-duration)]">
+              {readingTime ? (
+                // TODO: Use space-600 from the new colors when GDS is updated
```
benface, line 716: I feel like we'll have to replace almost every color class, but thanks for adding these TODOs!

website/src/components/CalendarIcon.tsx
benface, line 1: Very cool! Shouldn't we add it to GDS?

website/src/components/CalendarIcon.tsx
```diff
@@ -0,0 +1,33 @@
+import type { ComponentPropsWithoutRef } from 'react'
+
+interface CalendarIconProps extends ComponentPropsWithoutRef<'svg'> {
+  value?: number
```
benface, line 4: I think `day` or date` would be a better name?

website/src/components/CalendarIcon.tsx
```diff
@@ -0,0 +1,33 @@
...
+}
+
+export const CalendarIcon = ({ value = 1 }: CalendarIconProps) => {
```
benface, line 7: What if the default was the current date? Haha

website/src/components/CalendarIcon.tsx
```diff
@@ -0,0 +1,33 @@
+import type { ComponentPropsWithoutRef } from 'react'
+
+interface CalendarIconProps extends ComponentPropsWithoutRef<'svg'> {
```
benface, line 3: We should extend `ExperimentalIconProps` (or `WithOptional<ExperimentalIconProps, 'alt'>`) to support the `size` and `color` (CSS) props, and wrap the `<svg>` in `<ExperimentalIcon {...props}>` to make sure it works well everywhere icons are accepted. It does a bunch of helpful things, e.g. it sets `shrink-0` which ensures the icon doesn't get squished in a flex container with wrapping text, it is properly recognized as an icon in components that need to have different styles when they only contain an icon (e.g. `ExperimentalButton`), etc.

website/src/components/HourglassIcon.tsx
```diff
@@ -0,0 +1,37 @@
+import type { ComponentPropsWithoutRef } from 'react'
+
+interface HourglassIconProps extends ComponentPropsWithoutRef<'svg'> {
+  value?: number
```
benface, line 4: `fillPercent`?

website/src/components/HourglassIcon.tsx
```diff
@@ -0,0 +1,37 @@
...
+        height="14"
+      >
+        <path
+          d="M12.5 4.7275V2.5C12.5 2.23478 12.3946 1.98043 12.2071 1.79289C12.0196 1.60536 11.7652 1.5 11.5 1.5H4.5C4.23478 1.5 3.98043 1.60536 3.79289 1.79289C3.60536 1.98043 3.5 2.23478 3.5 2.5V4.75C3.50046 4.90523 3.53687 5.05824 3.60638 5.19704C3.67589 5.33584 3.7766 5.45665 3.90062 5.55L7.16687 8L3.9 10.45C3.77605 10.5434 3.67543 10.6642 3.60603 10.803C3.53663 10.9418 3.50034 11.0948 3.5 11.25V13.5C3.5 13.7652 3.60536 14.0196 3.79289 14.2071C3.98043 14.3946 4.23478 14.5 4.5 14.5H11.5C11.7652 14.5 12.0196 14.3946 12.2071 14.2071C12.3946 14.0196 12.5 13.7652 12.5 13.5V11.2725C12.5 11.1178 12.4641 10.9652 12.3951 10.8267C12.3262 10.6882 12.226 10.5676 12.1025 10.4744L8.82875 8L12.1025 5.52563C12.2258 5.43224 12.3258 5.31161 12.3948 5.17315C12.4638 5.0347 12.4998 4.88218 12.5 4.7275Z"
+          fill="white"
+        />
+      </mask>
+
+      <g mask="url(#hourglass-mask)">
+        <rect x="3.5" y={15 - value} width="9" height={value} fill="currentColor" />
```
benface, line 26: What happens if `value` is a ridiculously huge number? :P

website/src/Layout.tsx
```diff
@@ -784,7 +772,26 @@ function MDXContent({ toc: headings, children }: ComponentPropsWithoutRef<Nextra
           >
             {children}
           </section>
-          <div className="col-[container] flex justify-end pb-12 group-data-[hide-table-of-contents]/layout-toc-grid:hidden xl:hidden">
```
benface, line 787: Did you intentionally remove the `group-data-[hide-table-of-contents]/layout-toc-grid:hidden` class?

benface: Ah, sorry, I misunderstood what this was. This is fine :)

website/src/components/CalendarIcon.tsx
0xa3k5: yeah we can

website/src/components/CalendarIcon.tsx
0xa3k5: i think its a pretty solid idea

website/src/components/HourglassIcon.tsx
```diff
@@ -0,0 +1,40 @@
+import type { WithOptional } from '@edgeandnode/common'
+import { ExperimentalIcon as Icon, type ExperimentalIconProps as IconProps } from '@edgeandnode/gds'
+
+interface HourglassIconProps extends WithOptional<IconProps, 'alt'> {
+  fillPercent?: number
```
benface, line 5: @0xa3k5 – Thanks for changing the name, but the functionality should also change to work as a percentage. :P I can do that.

benface: But why 15? That seems like a completely arbitrary number. 🤔

0xa3k5: it is completely arbitrary, i just made it up. we don't have many documents where the reading time is >15 so its a easy enough general grouping

benface: Ok so it sounds like this value should be the reading time in minutes? I don't think this component should know about that at all, especially if we move it to GDS. That logic (converting the reading time to icon fill %) should be in the consumer.

benface: Ah, well I think both of them could potentially be used elsewhere (like in the blog – I know I always give that example but docs and blog are similar in a few ways haha). And even if it wasn't moved, I feel like accepting a fill % would be better code, but if you disagree, at the very least the prop name should say what it is, so `readingTime` (with a comment that it's in minutes).

0xa3k5: i didnt mean that to be a very precise indication, its just something extra that vast majority of the people will not even notice its there. feel free to make any necessary changes as you deem fit on the pr

benface: Ok, sorry for being nitpicky, I just thought `value` was unclear and then 15 appeared kind of a magic number in the component. I will move these to GDS and update your PR, and maybe we should also use the new Button instead of Link for the "Edit on GitHub" link like you said?

0xa3k5: agree but i feel like they should not be blockers for this PR, this is ready to be used, could we merge it?

benface: Agreed, let's just change this name to `readingTime`, add `TODO`s for moving the icons to GDS, and merge?

benface (approved):
Thank you so much!

PR #910, opened 2025-03-31
Adding Token API FAQ

MichaelMacaulay:


benface (commented):
Hey @MichaelMacaulay, did you run the `fix-pages-structure` script (or run `pnpm run build` which runs that script on pre-build) by any chance? Because we should avoid duplicating changes in all languages in PRs; it makes it harder / more confusing to review and address feedback.

benface (changes requested):
Also, looks like some images referenced in MDX files are missing, making the build fail.

CleanShot 2025-03-31 at 12 56 03@2x

MichaelMacaulay:
@benface yes, I ran fix-pages-structure

benface:
Ah ok, you shouldn't run that manually except maybe right before merging a PR, but even then, we can just run it in a separate branch / PR at some point. That script is run manually before build in CI/CD so not running it will not make the build / deployment fail.

PR #913, opened 2025-04-02
Supported networks pages

hayderkg:
### Supported networks:
- Search by network
- Testnet filter
- Network logos
- Token API column
- New design
- Dedicated pages per network
- Guide cards filtered by EVM vs Non-EVM


website/src/components/NetworkDetailsPage.tsx
```diff
@@ -0,0 +1,213 @@
...
+    'solana-accounts',
+    'solana-devnet',
+    'solana-mainnet',
+    'solana-mainnet-beta',
+    'solana-testnet',
+    'starknet-mainnet',
+    'starknet-testnet',
+  ]
+  return nonEVMIds.includes(networkId)
+}
```
0xa3k5, line 58: doesn't this create a maintenance burden? every time there is a new network in the networks-registry, we'll need to update here. Instead, we could rely on the `caip2id` [field](https://github.com/graphprotocol/networks-registry/blob/05cf0cbc816714e89158cdcd1588353fe56a1f7a/schemas/registry.schema.json#L61). Caip2Id for the EVM networks starts with `eip155` others don't.

website/src/components/NetworkDetailsPage.tsx
```diff
@@ -0,0 +1,213 @@
...
+    <div className={classNames(['flex items-center gap-1 leading-none', className])} {...props}>
+      <Clock
+        alt={variant === 'reading' ? t('index.time.reading') : t('index.time.duration')}
+        variant="fill"
+        size={3.5}
+      />
+      {minutes} {t('index.time.minutes')}
+    </div>
+  )
+}
```
0xa3k5, line 80: this should be replaced with [`HourglassDynamic`](https://github.com/edgeandnode/components/blob/main/packages/gds/src/experimental/icons/custom/HourglassDynamic.tsx)  like the other cards in the docs, sorry if the Figma was outdated for these. there is a local component [`TimeIcon`](https://github.com/graphprotocol/docs/blob/main/website/src/components/TimeIcon.tsx)

website/src/components/NetworkDetailsPage.tsx
```diff
@@ -0,0 +1,213 @@
...
+        <Card
+          href="https://docs.substreams.dev/"
+          title={t('index.networkGuides.nonEvm.officialDocs.title')}
+          description={t('index.networkGuides.nonEvm.officialDocs.description')}
+          slotAboveTitle={<Time variant="reading" minutes={15} />}
+          className="min-h-[252px]"
+          icon={
+            <div className="flex size-8 items-center justify-center text-white">
+              <Substreams size={6} />
+            </div>
```
0xa3k5, line 158: i'm a bit confused with this, curious why do we need to wrap the icon with a bigger div? We probably should just remove these divs and just pass the icon

website/src/supportedNetworks.tsx
```diff
@@ -1,17 +1,71 @@
...
+
+// Skeleton networks (no icon available)
+export const MISSING_ICON_NETWORKS = [
+  'berachain-bepolia',
+  'hoodi',
+  'ink-sepolia',
+  'megaeth-testnet',
+  'solana-accounts',
+  'manta',
+]
```
0xa3k5, line 54: all of these except `hoodi` has web3icon, i can add `hoodi` pretty soon. 

https://github.com/0xa3k5/web3icons/blob/main/docs/icons.md

the testnet names, might be the why you couldn't find the icons but the icon names are correctly identified in the `networks-registry`'s `icon.web3icon` field ([example](https://github.com/graphprotocol/networks-registry/blob/05cf0cbc816714e89158cdcd1588353fe56a1f7a/registry/eip155/berachain-bepolia.json#L33))

website/src/supportedNetworks.tsx
```diff
@@ -1,17 +1,71 @@
...
+
 export async function getSupportedNetworks() {
   const registry = await NetworksRegistry.fromLatestVersion()
 
+  // Networks with Token API support
+  const tokenAPINetworks = ['mainnet', 'base', 'bsc', 'arbitrum-one', 'matic', 'optimism']
```
0xa3k5, line 60: shouldn't we be relying on the `services` field for these? https://github.com/graphprotocol/networks-registry/blob/05cf0cbc816714e89158cdcd1588353fe56a1f7a/registry/eip155/mainnet.json#L25

0xa3k5: interestingly, i can't see anything for `token-api` maybe we should ask someone to update the registry (Yaro maybe?)

website/src/supportedNetworks.tsx
```diff
@@ -34,33 +89,323 @@ export async function getSupportedNetworksStaticProps() {
...
+        <ExperimentalToggleChip
+          size="small"
+          onChange={handleToggleTestnets}
+          checked={showTestnets}
+          disabled={isLoadingTestnets}
+          iconBefore={isLoadingTestnets ? <LoadingSpinner className="size-4" /> : undefined}
+        >
+          {t('index.supportedNetworks.showTestnets')}
+        </ExperimentalToggleChip>
+      </div>
```
0xa3k5, line 294:
```suggestion
        <ExperimentalToggleChip
          size="small"
          onChange={handleToggleTestnets}
          checked={showTestnets}
        >
          {t('index.supportedNetworks.showTestnets')}
        </ExperimentalToggleChip>
      </div>
```

One of the biggest design decisions that we made while designing these new Chip components was to ensure that their length always stays the same, no layout shifts, no other content pushes. we tried very hard to achieve this. And now, I see with this approach, the Search bar is being pushed because of the loading state. 

I seriously don't believe we need a loading state, this is almost instant (and there is no server/api calls to filter out the list) so I tried in my local how it feels without this loading state, and i feels good. Can we please remove the loading states here? 

https://share.cleanshot.com/Jt0mPrcT

website/src/supportedNetworks.tsx
0xa3k5: just realized `hoodi` is a testnet of `ethereum mainnet` so we should just use the `ethereum` icon

0xa3k5 (changes requested):

0xa3k5:
good work @hayderkg , just left a few comments

website/src/components/NetworkDetailsPage.tsx
hayderkg: will do! thanks

website/src/supportedNetworks.tsx
hayderkg: I think they are still working on it as BK told me recently

website/src/components/NetworkDetailsPage.tsx
hayderkg: You are totally right.

website/src/supportedNetworks.tsx
hayderkg: I think this requires to update the componente NetworkIcon from GDS. Will do it.

benface: I already have most of these updated locally, I can create a PR in a bit.

benface: https://github.com/edgeandnode/components/pull/1332

package.json
```diff
@@ -35,5 +35,8 @@
   },
   "resolutions": {
     "@babel/core": "7.26.7"
+  },
+  "dependencies": {
+    "framer-motion": "^12.6.3"
```
benface, line 40: This should be added to the `website` package. Notice how all the dependencies in this file are the shared dev ones.

website/src/HomePage.tsx
```diff
@@ -259,11 +259,11 @@ function SupportedNetworks({ className, ...props }: ComponentPropsWithoutRef<'di
       <ul className="grid grid-cols-auto-fill-16 gap-px text-space-500">
         {supportedNetworks
           // TODO: Don't filter out testnets that don't have a mainnet
-          .filter((network) => network.networkType === NetworkType.Mainnet)
+          .filter((network) => String(network.networkType) === String(NetworkType.Mainnet))
           // Filter out networks that are either duplicates (same logo, same or similar short name) or irrelevant in this view
           .filter(
             (network) =>
-              !network.caip2Id.startsWith('beacon:') &&
+              !network.caip2Id?.startsWith('beacon:') &&
```
benface, line 266: `caip2Id` should always be set 🤔

CleanShot 2025-04-02 at 10 25 38@2x

website/src/components/NetworkDetailsPage.tsx
```diff
@@ -0,0 +1,138 @@
...
+
+  return (
+    <div className="graph-docs-not-markdown mt-8 grid grid-cols-1 gap-4">
+      <div className="grid grid-cols-1 gap-4 sm:grid-cols-2 lg:grid-cols-3">
+        <Card
+          href="https://thegraph.com/docs/en/subgraphs/quick-start/"
+          title={t('index.networkGuides.evm.subgraphQuickStart.title')}
+          description={t('index.networkGuides.evm.subgraphQuickStart.description')}
+          slotAboveTitle={<TimeIcon variant="reading" minutes={10} />}
+          className="min-h-[252px]"
```
benface, line 35: Curious why this is necessary?

website/src/components/NetworkDetailsPage.tsx
```diff
@@ -0,0 +1,138 @@
...
+        />
+      </div>
+    </div>
+  )
+})
+
+EVMResources.displayName = 'EVMResources'
+NonEVMResources.displayName = 'NonEVMResources'
+
+const NetworkDetailsPage = memo(({ network }: NetworkDetailsPageProps) => {
```
benface, line 125: Interesting, I never use `memo`. Does it actually help performance here?

website/src/components/NetworkPage.tsx
```diff
@@ -0,0 +1,137 @@
...
+        <NetworkDetailsPage
+          network={{
+            id: networkData.id,
+            fullName: networkData.fullName,
+            networkType: networkData.networkType,
+            protocol: networkData.graphNode?.protocol ?? '',
+            chainId: networkData.caip2Id?.split(':')[1] ?? '',
+            nativeCurrency: networkData.nativeToken ?? '',
+            docs: networkData.docsUrl ?? '',
+          }}
```
benface, line 130: Why this mapping, instead of passing `networkData` as is?

benface: I feel like the renaming of some properties (e.g. `caip2Id` to `chainId` and `nativeToken` to `nativeCurrency`) is a source of potential confusion.

website/src/components/NetworkPage.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+  return (
+    <>
+      <Head>
+        <title>{networkData.fullName}</title>
+      </Head>
+      <div className="col-[container]">
+        <div className="mb-5 mt-12 flex flex-col gap-3">
+          {networkData.caip2Id ? (
+            <div className="flex-shrink-0">
```
benface, line 53: `flex-shrink-0` is a deprecated Tailwind class, and has even [been removed in Tailwind 4](https://tailwindcss.com/docs/upgrade-guide#removed-deprecated-utilities) (the correct one is just `shrink-0`). [This rule](https://github.com/francoismassart/eslint-plugin-tailwindcss/blob/master/docs/rules/migration-from-tailwind-2.md) from `eslint-plugin-tailwindcss` (which we use) would catch it, but unfortunately [it also catches false positives](https://github.com/francoismassart/eslint-plugin-tailwindcss/issues/294) and so it is currently unusable. I am monitoring future updates to `eslint-plugin-tailwindcss` to include it in a future release of `@edgeandnode/eslint-config` (which will be based on the ESLint 9 setup in `@edgeandnode/components`); someone even has a PR open for Tailwind 4 support... All this to say, for now, let's remember to use `shrink-*`/`grow-*` instead of `flex-shrink-*`/`flex-grow-*`, but I'll do my best to automate that soon.

website/src/utils/networkUtils.ts
```diff
@@ -0,0 +1,139 @@
...
+  'linea-sepolia',
+  'linea',
+  'lens-testnet',
+  'expchain-testnet',
+  'autonomys-taurus',
+  'arweave-mainnet',
+  'fraxtal',
+  'lumia',
+  'mbase',
+]
```
benface, line 21: Nice! Ideally this is not needed in the future, but in the meantime, this is a neat solution.

website/src/utils/networkUtils.ts
```diff
@@ -0,0 +1,139 @@
...
+
+// Skeleton networks (no icon available)
+export const MISSING_ICON_NETWORKS = [
+  'berachain-bepolia',
+  'hoodi',
+  'ink-sepolia',
+  'megaeth-testnet',
+  'solana-accounts',
+  'manta',
+]
```
benface, line 31: None of these will be missing anymore after the GDS update, but if we want to keep this `MISSING_ICON_NETWORKS` array, I would add a TODO that we should handle it in GDS/GO somehow instead.

website/src/utils/networkUtils.ts
```diff
@@ -0,0 +1,139 @@
...
+
+// Non-EVM networks
+export const NON_EVM_NETWORKS = [
```
benface, line 34: Like AK said, I am pretty sure we have this information in the registry already, or can infer it based on something more general than specific network names/IDs?

website/src/utils/networkUtils.ts
```diff
@@ -0,0 +1,139 @@
...
+  'solana-devnet',
+  'solana-mainnet',
+  'solana-mainnet-beta',
+  'solana-testnet',
+  'starknet-mainnet',
+  'starknet-testnet',
+]
+
+// Networks with Token API support
+export const TOKEN_API_NETWORKS = ['mainnet', 'base', 'bsc', 'arbitrum-one', 'matic', 'optimism']
```
benface, line 66:
```suggestion
// Networks with Token API support (TODO: remove once the registry has this information)
export const TOKEN_API_NETWORKS = ['mainnet', 'base', 'bsc', 'arbitrum-one', 'matic', 'optimism']
```

website/src/components/NetworkPage.tsx
```diff
@@ -0,0 +1,137 @@
...
+                  size={10}
+                  className="h-10 w-10"
+                />
+              )}
+            </div>
+          ) : null}
+          <h2 className="leading-tight mt-0 text-24 text-white">{networkData.fullName}</h2>
+        </div>
+
+        <Grid className="gap-4">
```
benface, line 69: This is an old component, we should really mark it as deprecated in GDS (I guess it will be once I get to [this task](https://linear.app/thegraph/issue/GDS-298/update-file-structure-and-deprecatepromote-components)). The new way is to use Tailwind's `grid` class on a `div` (or any element).

benface: (Same for `Flex` and `Text`)

website/src/components/index.ts
```diff
@@ -5,6 +5,9 @@ export * from './DocSearch'
 export * from './Heading'
 export * from './Image'
 export * from './Navigation'
+export * from './NetworkDetailsPage'
+export * from './NetworkPage'
+export * from './supported-networks'
```
benface, line 10: I would prefer if `components` remained about "global" components (those that can be used on any MDX page). I would personally move all of these to a `supportedNetworks` directory.

website/src/components/supported-networks/EmptySearchResults.tsx
```diff
@@ -0,0 +1,42 @@
...
+      <div className="text-space-700">
+        <DottedRingsSpinner>
+          <EyeClosed
+            size="36px"
+            color="purple-100"
+            alt="Information"
+            sx={{
+              stroke: 'purple-100',
+              transform: 'translate(1px, -1.5px)',
+            }}
```
benface, line 27: Noooo please no Theme UI / Emotion in new code. 😭

website/src/components/supported-networks/EmptySearchResults.tsx
```diff
@@ -0,0 +1,42 @@
...
+          />
+        </DottedRingsSpinner>
+      </div>
+      <Text.P20 className="mb-2 mt-0 font-medium">{t('index.supportedNetworks.emptySearch.title')}</Text.P20>
```
benface, line 31: You should use the `text-p20` class instead of the `<Text.P20>` component, or even better: the Tailwind theme for the docs defines more contextual / semantic text style classes, and in this case, `text-p20` maps to `text-body-large`. (Typography still needs to be cleaned up in GDS BTW; I believe @0xa3k5 and Edgar are working on it.)

website/src/components/supported-networks/EmptySearchResults.tsx
```diff
@@ -0,0 +1,42 @@
...
+      <Text.P16 className="mb-6 max-w-md text-space-700">
```
benface, line 32: Hmm, I didn't realize we kept the "named" `max-w` classes in our Tailwind theme. I'm not a fan of those because they are quite opinionated (and not our opinions, i.e. I have no idea what "md" is). I would prefer using a number.

website/src/supportedNetworks.tsx
```diff
@@ -38,29 +27,123 @@ export function SupportedNetworksTable({
...
+    if (noResults) {
+      setSearchQuery('')
+    }
+
+    setIsLoadingTestnets(true)
+
+    setTimeout(() => {
+      setShowTestnets(true)
+      setIsLoadingTestnets(false)
+    }, 150)
```
benface, line 94: Not sure what is the point of this artificial delay?

website/src/supportedNetworks.tsx
```diff
@@ -38,29 +27,123 @@ export function SupportedNetworksTable({
...
+          network.fullName.toLowerCase().includes(query),
+      )
+    }
+
+    return filtered
+  }, [networks, searchQuery, showTestnets])
+
+  useEffect(() => {
+    setNoResults(filteredNetworks.length === 0 && searchQuery !== '')
+  }, [filteredNetworks.length, searchQuery])
```
benface, line 78: I believe `noResults` doesn't need to be a state variable at all, and can just be set synchronously during render instead of in a `useEffect`.

website/src/supportedNetworks.tsx
```diff
@@ -38,29 +27,123 @@ export function SupportedNetworksTable({
...
-            <td align="center">{network.subgraphs ? '✓' : null}</td>
-            <td align="center">{network.substreams ? '✓' : null}</td>
-            <td align="center">{network.firehose ? '✓' : null}</td>
-          </tr>
-        ))}
-      </tbody>
-    </Table>
+    <>
+      <Callout variant="info" className="mb-6 flex items-center">
+        <Text.P14>
```
benface, line 118: Should be `text-p14`, or `text-body-xsmall`, or... nothing at all? `Callout` already sets a font size (though it's `text-body-small` instead of `text-body-xsmall`, but we should be consistent, right?)

website/src/components/supported-networks/EmptySearchResults.tsx
benface: This is super nitpicky BTW, feel free to ignore. It was mostly a comment for myself (I regret not removing them from the theme in GDS but now it would be a breaking change to do so, heh).

website/src/utils/networkUtils.ts
benface, line 1: I'm a big fan of colocation. What do you think about moving this file to a `src/supportedNetworks` directory along with `supportedNetworks.tsx` (potentially renaming / splitting that) and the related components as mentioned [here](https://github.com/graphprotocol/docs/pull/913/files#r2025018472)?

website/src/components/NetworkDetailsPage.tsx
hayderkg: Because in the design there is a fixed height for the container of 252px, I suppose AK wants to avoid little space between the logo and the title but enough for large texts.
image

website/src/components/NetworkPage.tsx
```diff
@@ -0,0 +1,137 @@
...
+                  <ExperimentalDescriptionList.Item label={t('index.supportedNetworks.protocol')}>
+                    {networkData.graphNode.protocol}
+                  </ExperimentalDescriptionList.Item>
+                )}
+                {networkData.id && (
+                  <ExperimentalDescriptionList.Item variant="mono" label={t('index.supportedNetworks.identifier')}>
+                    <div className="flex items-center gap-2">
+                      {networkData.id}
+                      <ExperimentalCopyButton size="small" variant="naked" value={networkData.id} />
+                    </div>
```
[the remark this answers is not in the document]
hayderkg: Awesome, thank you.

website/src/components/NetworkPage.tsx
hayderkg: Awesome, thanks.

website/src/utils/networkUtils.ts
hayderkg: Good one, thanks.

website/src/components/Table.tsx
```diff
@@ -1,10 +1,14 @@
 import type { ComponentPropsWithoutRef } from 'react'
 
-import { classNames, Text } from '@edgeandnode/gds'
+import { classNames } from '@edgeandnode/gds'
 
-interface TableProps extends ComponentPropsWithoutRef<'table'> {}
+interface TableProps extends ComponentPropsWithoutRef<'table'> {
+  variant?: 'default' | 'supported-networks'
+}
```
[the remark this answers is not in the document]
hayderkg: Wasn't sure about this… shall we merge as it is and then talk about updating the whole component?

website/src/components/index.ts
hayderkg: Oh yes, for sure

website/src/components/Table.tsx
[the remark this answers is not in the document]
benface: Sure!

website/src/components/supported-networks/EmptySearchResults.tsx
hayderkg: Sorry... copy/pasted this from GDS 😅🫣

benface: Haha... sorry for my exaggerated reaction. 🙇‍♂️

website/src/components/supported-networks/EmptySearchResults.tsx
hayderkg: Thank you

website/src/components/supported-networks/EmptySearchResults.tsx
hayderkg: Oh no, I find it very interesting. Thanks.

website/src/supportedNetworks.tsx
hayderkg: Oh, it’s to improve the UX. Before, we had a spinner—normally it loads instantly, but sometimes it takes around 100ms to load. If I don’t add a delay—which, as you can see, is minimal—you just see a “flash” and the spinner isn’t even noticeable. So with just an extra 50ms (roughly), we get the best of both worlds.

website/src/supportedNetworks.tsx
hayderkg: Yes, if they're already available, there's no need. Good catch.

website/src/supportedNetworks.tsx
```diff
@@ -38,29 +27,123 @@ export function SupportedNetworksTable({
...
+    } else {
+      setIsLoading(true)
+
+      const timer = setTimeout(() => {
+        setIsLoading(false)
+      }, 1500)
+
+      return () => clearTimeout(timer)
+    }
+  }, [networks])
```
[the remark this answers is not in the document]
hayderkg: It was due to the loading status, but it seems that it is not necessary.

website/src/supportedNetworks.tsx
hayderkg: Removing the component Text from here but I'm not sure if we want to have it as `text-body-small`:
image

I will make it `xsmall` for now.

website/src/utils/networkUtils.ts
hayderkg: Great idea.
Did: src/supportedNetworks with `index.tsx` + `utils.ts` + /components

benface: Yes! 🙌 `index.tsx` should probably just be a barrel file; `getSupportedNetworks` and `getSupportedNetworksStaticProps` could be moved to `utils.ts` and `SupportedNetworksTable` could be extracted into its own `SupportedNetworksTable.tsx` file.

website/src/components/NetworkDetailsPage.tsx
benface: Ok, I wonder if that was intentional from @0xa3k5 (as I think the same style of `Card` looks fine on the home page without this min height).

website/src/supportedNetworks.tsx
benface: (I have a `useEffect` paranoia, I feel like 90% of them can be avoided, resulting in more performant and often less buggy code.)

website/src/supportedNetworks.tsx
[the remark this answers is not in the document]
benface: Yeah, since we're never fetching the networks on the client-side.

website/src/supportedNetworks.tsx
benface: Huh, you're right; looks like I made a mistake implementing [the `Callout` component](https://www.figma.com/design/ZHcIcsuiOoNgsLXRsa8KPk/Docs?node-id=1494-34913&t=yxpto5cL2SnueF88-1). So I would change `text-body-small` to `text-body-xsmall` in `Callout`. Also, not sure what's going on with the icon here; it should only be centered with the first line. 🤔

benface: Oh, that's because you added `items-center` to `<Callout>`! I would remove that, and `flex` too; those declarations break the API boundary of the component IMO (what if `Callout` changed and its root element was no longer a flex container?)

benface: I always go back to [this tweet by Mark Dalgleish](https://x.com/markdalgleish/status/1338416121242796033) that I find "so true". Though I would nuance it a bit: I believe that _some_ properties should (as in, we should ensure that they are) safe to use on most if not all components; I'm talking about properties that don't affect the internal styles of the component at all, but rather its external layout: `position`, `margin`, `z-index`, `order`, `flex-grow/shrink`, etc. There's also more tricky nuance around custom properties; some of them are designed to be public, while others to be private... but I think the general idea of being careful not to break the "CSS encapsulation" of a component is a good one, as it is almost impossible to enforce due to how CSS works (of course, we could just not accept `style` or `className`, but that would be too strict; it would require wrapping every component that we need to add margin to in a div or something... not practical). One day I'll write an ESLint plugin for that or something. 😄

website/src/components/NetworkDetailsPage.tsx
0xa3k5: home page looks fine yes, but this is slightly more preferable

benface: Cool. But since 252px is divisible by 4, I would at least change this to `min-h-63` @hayderkg :P

0xa3k5: no no, let's please prefer 256 over 252 😅

0xa3k5: min-h-64h 🤝

website/src/components/NetworkPage.tsx
0xa3k5: I guess this is mainly because of how we have it in the designs, we don't have `caip2id` but was meant to display the `chainId` as it's a way more common (at least among the EVM networks) way of identifying the network. however, I agree, if we are going to display `caip2id` we should not label it as `chain id`. I really thought we would have `chainId` field in the registry but apparently not, so maybe change the label to `caip2id` as well? 

alternatively, not sure how I feel about this, we can display `chainId` on the EVM networks (by removing the `eip155:` from the `caip2id`) and `caip2id` in the other networks. I don't hate this at all, just slightly annoying because chain id is mostly a evm thing.

for the native token, this is very annoying to me, most other "standard" libraries name this `nativeCurrency` but turns out the registry names it `nativeToken` so I guess we should just call it `Native Token`

website/src/HomePage.tsx
```diff
@@ -259,11 +259,11 @@ function SupportedNetworks({ className, ...props }: ComponentPropsWithoutRef<'di
       <ul className="grid grid-cols-auto-fill-16 gap-px text-space-500">
         {supportedNetworks
           // TODO: Don't filter out testnets that don't have a mainnet
-          .filter((network) => network.networkType === NetworkType.Mainnet)
+          .filter((network) => String(network.networkType) === String(NetworkType.Mainnet))
```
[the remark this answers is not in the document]
hayderkg: Sorry, pretty sure it was failing. Fixed.

website/src/components/NetworkPage.tsx
0xa3k5: oh just realized we are splitting the caip2id, that is not going to work well for the non-evm networks, i'm more convinced we should just display the `caip2id` instead. (maybe we can actually help make the `caip2id` mainstream)

website/src/components/NetworkDetailsPage.tsx
hayderkg: Yep. Maybe not very significant here but it just avoid re-renders, only if {network} changes.

website/src/components/NetworkPage.tsx
hayderkg: It was just because I was extracting the chainId from the caip2Id and renaming the properties. Removed

website/src/components/NetworkDetailsPage.tsx
hayderkg: Excellent. Thank you!

website/src/supportedNetworks/utils.ts
```diff
@@ -25,38 +27,6 @@ export const MONO_ICON_NETWORKS = [
 // Skeleton networks (no icon available)
 export const MISSING_ICON_NETWORKS = ['ink-sepolia']
```
0xa3k5, line 28: ink is available in the web3icons, is this GDS issue?

benface: Yes, looks like a GDS issue. My bad, PR here: https://github.com/edgeandnode/components/pull/1334

website/package.json
```diff
@@ -17,19 +17,20 @@
...
+    "@edgeandnode/gds": "^6.5.12",
+    "@edgeandnode/go": "^9.4.12",
     "@emotion/react": "^11.14.0",
     "@graphprotocol/contracts": "6.2.1",
     "@pinax/graph-networks-registry": "^0.6.7",
     "@react-hookz/web": "^25.1.0",
     "@readme/httpsnippet": "^11.0.0",
     "@readme/openapi-parser": "^3.0.1",
     "fetch-har": "^11.1.1",
+    "framer-motion": "^12.6.3",
```
benface, line 30: @hayderkg – Look at line 33. Motion was already installed (under its new name), so it's now duplicated. 😛

website/src/pages/[locale]/supported-networks/[id].mdx
```diff
@@ -0,0 +1,65 @@
...
+export const getStaticPaths = async () => {
+  const networks = await getSupportedNetworks()
+  const paths = []
+  for (const locale of supportedLocales) {
+    for (const network of networks) {
+      paths.push({
+        params: { locale, id: network.id },
+      })
+    }
+  }
```
Comment (benface) [inline website/src/pages/[locale]/supported-networks/[id].mdx:18]: This is super readable and nice, but nitpick: initializing `paths` this way will make its type `any[]`. I'm actually surprised that not only this doesn't emit a warning by our `typescript-eslint` rules, but [there doesn't seem to be a rule for that](https://stackoverflow.com/questions/77150939/triggering-typescript-error-for-untyped-arrays)!? Seems really insidious.

website/src/supportedNetworks/components/NetworkPage.tsx
```diff
@@ -0,0 +1,116 @@
...
+                    {networkData.nativeToken}
+                  </ExperimentalDescriptionList.Item>
+                )}
+                {networkData.docsUrl && (
+                  <ExperimentalDescriptionList.Item label={t('index.supportedNetworks.docs')}>
+                    <ExperimentalLink
+                      className="text-p14"
+                      href={networkData.docsUrl}
+                      target="_blank"
+                      rel="noopener noreferrer"
```
benface, line 97: This `rel` is not needed, it's handled automatically by `ExperimentalLink` (or more precisely by the underlying `ButtonOrLink`).

website/src/supportedNetworks/components/NetworkRow.tsx
```diff
@@ -0,0 +1,79 @@
...
+    substreams: boolean
+    firehose: boolean
+    tokenapi: boolean
+  }
+  locale: string
+}
+
+export const NetworkRow = memo(({ network, locale }: NetworkRowProps) => {
+  return (
+    <NextLink href={`/${locale}/supported-networks/${network.id}`} legacyBehavior passHref>
```
benface, line 28: Huh, does this `legacyBehavior` mean that this is not actually rendering an anchor? If so, I believe it's a problem for accessibility. Can you still tab through the rows?

website/src/supportedNetworks/components/NetworkRow.tsx
```diff
@@ -0,0 +1,79 @@
...
+              <Skeleton borderRadius="FULL" height="20px" width="20px" />
+            ) : (
+              <NetworkIcon variant={getIconVariant(network.id)} caip2Id={network.caip2Id as any} size={5} />
+            )}
+            <Text.P14>{network.shortName}</Text.P14>
+          </div>
+        </td>
+        <td className="w-48">
+          <div className="flex w-full items-center justify-between gap-2">
+            <Text.P14 className="!mb-0">{network.id}</Text.P14>
```
benface, line 59: No big deal, but there are still some instances of `Text` in this file.

website/src/supportedNetworks/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+    if (!showTestnets) {
+      handleShowTestnets()
+    } else {
+      handleHideTestnets()
+    }
+  }, [showTestnets, handleShowTestnets, handleHideTestnets])
+
+  const setSearchQueryCallback = useCallback((value: string) => {
+    setSearchQuery(value)
+  }, [])
```
benface, line 85: These `useCallback`s don't seem like they will improve performance: https://kentcdodds.com/blog/usememo-and-usecallback

benface: (I would just inline all of these handlers where they are used below personally.)

website/src/supportedNetworks/utils.ts
benface: Just merged; when you see this you'll be able to update GDS in this branch @hayderkg. 🙏 Sorry about that!

website/src/components/NetworkPage.tsx
benface: Weird, "native token" makes more sense to me, but looks like Hayder still used "Native Currency" in the UI @0xa3k5.

website/src/supportedNetworks/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+    <>
+      <Callout variant="info" className="mb-6 flex items-center">
+        <span className="text-p14">
+          {t('index.supportedNetworks.infoText')}{' '}
+          <ExperimentalLink href="https://thegraph.com/docs/en/indexing/chain-integration-overview/">
+            {t('index.supportedNetworks.infoLink')}
+          </ExperimentalLink>
+          .
+        </span>
+      </Callout>
```
benface, line 97:
```suggestion
      <Callout variant="info" className="mb-6">
        {t('index.supportedNetworks.infoText')}{' '}
        <ExperimentalLink href="https://thegraph.com/docs/en/indexing/chain-integration-overview/">
          {t('index.supportedNetworks.infoLink')}
        </ExperimentalLink>
        .
      </Callout>
```
(as discussed in [this thread](https://github.com/graphprotocol/docs/pull/913#discussion_r2025049820))

benface: (Feel free to update `Callout` to fix the font size, sorry I can't make the suggestion here because that file hasn't been updated obviously.)

benface: Oh, and also, related to the separation of `infoText` and `infoLink`: you can actually use a single translation string for that (and I would also include the period at the end in it), and I believe it's better because some languages may want to translate the sentence in a way that the link is not at the very end. The way to do that is to use a `{0}` placeholder for the link, and pass an array containing the string or `ReactNode` to replace it with as the second argument to `t()`. Check out `index.supportedNetworks.description.base` (in this same codebase) to see what I mean!

benface: Though... I just remembered that `infoLink` is a self-contained CTA in this case, and not the continuation of a sentence like the code looks like. So I guess it's not super important here.

website/src/supportedNetworks/utils.ts
hayderkg: Awesome. Thank you.

website/package.json
hayderkg: It seems like a different package? it doesn't look like the official Framer one, why do we want to use this one and not framer-motion?

website/src/supportedNetworks/components/NetworkRow.tsx
```diff
@@ -0,0 +1,79 @@
...
+import NextLink from 'next/link'
+import { memo } from 'react'
+
+import { ExperimentalCopyButton, Skeleton, Text } from '@edgeandnode/gds'
+import { Check } from '@edgeandnode/gds/icons'
+import { NetworkIcon } from '@edgeandnode/go'
+
+import { getIconVariant, shouldShowSkeleton } from '@/supportedNetworks/utils'
+
+const MotionTR = motion.tr
```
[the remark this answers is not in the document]
hayderkg: Yeah, it seems the same ^^

website/src/supportedNetworks/components/NetworkRow.tsx
hayderkg: The problem was that it broke the table's UI, but I've made it absolutely positioned to handle this. I've also taken the opportunity to improve accessibility by displaying the CopyButton when you tab to the second tab.

https://github.com/user-attachments/assets/9c7c0429-e111-4fbd-a246-53f48ccaeff8

website/src/supportedNetworks/components/NetworkRow.tsx
hayderkg: Ouch! sorry…

website/src/supportedNetworks/index.tsx
hayderkg: Inlined the handlers + removed the callbacks. We can add them back if we notice performance issues or re-renders are problematic.

website/src/pages/[locale]/supported-networks/[id].mdx
Comment (hayderkg) [reply website/src/pages/[locale]/supported-networks/[id].mdx:18]: Using flatMap + map instead for loops

website/package.json
0xa3k5: motion has left Framer, and is now just `motion`

website/src/pages/[locale]/supported-networks/[id].mdx
Comment (benface) [reply website/src/pages/[locale]/supported-networks/[id].mdx]: You know what, I was completely wrong. `paths` is `any[]` when first defined, but that has the ability to change. TIL about "evolving any": https://effectivetypescript.com/2020/03/09/evolving-any/

But of course `flatMap` and `map` is also fine. Sorry for the wrong information, though!

website/package.json
hayderkg: No way! Sorry, didn't know this.

benface: Haha yep! Blog post here: https://motion.dev/blog/framer-motion-is-now-independent-introducing-motion

website/src/supportedNetworks/components/NetworksTable.tsx
```diff
@@ -0,0 +1,59 @@
...
+          </th>
+        </tr>
+        {isLoading ? (
+          skeletonRows
+        ) : (
+          <AnimatePresence initial={false}>
+            {networks.map((network) => (
+              <NetworkRow key={network.id} network={network} locale={locale} />
+            ))}
+          </AnimatePresence>
```
benface, line 54: Is this `AnimatePresence` for this animation? Seems a bit strange IMO.

https://github.com/user-attachments/assets/3fab3911-e392-4475-a0a0-3433c0953d60

benface: Also, I'm seeing really bad performance in Chrome. :( The above video was recorded in Zen/Firefox, but here's what it looks like in Chrome:


https://github.com/user-attachments/assets/dfa46762-a330-43da-a37a-3e4a6afdd61a

hayderkg: That's super weird, first time I see such a thing… (and I'm in Brave). Will remove it.

website/src/supportedNetworks/components/NetworkRow.tsx
benface: I made a PR to your PR with a fix for this: https://github.com/graphprotocol/docs/pull/918

0xa3k5 (approved):

benface (approved):
Beautiful. Ship it! 🚀

PR #923, opened 2025-04-08
Add resources for chain-specific landing pages

hayderkg:
- Extract feature detection into `getNetworkFeatures`
- Created a `getResourceCardsConfig` function to handle resource card selection
- Simplified main component
- Created `ResourceCards.tsx` to extract the renderig logic into `CardSection` and remove code duplication in the cards layouts


hayderkg (commented):
LGTM, but it's a bit verbose to repeat it in each case. Maybe we should keep a closer eye on it later.

benface (approved):

PR #939, opened 2025-04-29
Moving SpS

idalithb:
- Moving SpS under Substreams in the nav


website/src/pages/en/substreams/_meta.js
```diff
@@ -5,4 +5,5 @@ export default {
   introduction: '',
   developing: titles.developing,
   publishing: '',
+  sps: 'Substreams-Powered Subgraphs',
```
benface, line 8:
```suggestion
  sps: titles.sps,
```

benface (approved):
🙌

PR #967, opened 2025-05-29
Distributed systems explanation and how to guide

MichaelMacaulay:
Split Distributed systems into explanation page and how to guide

### DIFF
--- a/website/src/pages/en/subgraphs/querying/_meta.js
+++ b/website/src/pages/en/subgraphs/querying/_meta.js
@@ -6,7 +6,7 @@ export default {
   'best-practices': '',
   'from-an-application': '',
   'distributed-systems': '',
-  'distributed-systems-guide': '',
+  'distributed-systems-guide': titles['distributed-systems-guide'] ?? '',
   'graphql-api': '',
   'subgraph-id-vs-deployment-id': '',
   'graph-client': titles['graph-client'] ?? '',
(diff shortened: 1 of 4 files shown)


website/src/pages/en/subgraphs/querying/_meta.js
```diff
@@ -6,6 +6,7 @@ export default {
   'best-practices': '',
   'from-an-application': '',
   'distributed-systems': '',
+  'distributed-systems-guide': '',
```
benface, line 9:
```suggestion
  'distributed-systems-guide': titles['distributed-systems-guide'] ?? '',
```

website/src/pages/en/subgraphs/querying/distributed-systems-guide.mdx
```diff
@@ -0,0 +1,124 @@
...
+}
+```
+
+## Recap and Next Steps
+
+By using the `number_gte` parameter in a polling loop, you ensure time moves forward when fetching updates. By pinning queries to a specific `block.hash`, you can retrieve multiple sets of related information consistently from the same block.
+
+• If you encounter re-orgs, plan to retry from the beginning or adjust your logic accordingly. • Explore other filtering and block arguments (see \[placeholder for reference location\]) to handle additional use-cases.
+
+\[Placeholder for additional references or external resources if available\]
```
benface, line 124: These are weird (• characters and placeholders).

PR #971, opened 2025-06-03
Fixes for Indexing Section 

idalithb:
- Removing Agora 
- Changing language around graph-node

### DIFF
--- a/website/src/pages/en/indexing/tooling/graph-node.mdx
+++ b/website/src/pages/en/indexing/tooling/graph-node.mdx
@@ -89,7 +89,7 @@ When it is running Graph Node exposes the following ports:
 >
 > - Exposing Graph Node's internal ports can lead to a full system compromise.
 > - These ports must remain **private**: JSON-RPC Admin endpoint, Indexing Status API, and PostgreSQL.
-> - Do not expose 8000 (GraphQL HTTP) and 8001 (GraphQL WebSocket) directly to the internet. Even though these are used for GraphQL queries, they should ideally be proxied though `indexer-agent` and served behind a production-grade proxy. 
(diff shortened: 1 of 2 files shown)


jacordaro (approved):

website/src/pages/en/indexing/tooling/graph-node.mdx
```diff
@@ -85,7 +85,11 @@ When it is running Graph Node exposes the following ports:
 | 8030 | Subgraph indexing status API | /graphql | --index-node-port | - |
 | 8040 | Prometheus metrics | /metrics | --metrics-port | - |
 
-> **Important**: Be careful about exposing ports publicly - **administration ports** should be kept locked down. This includes the the Graph Node JSON-RPC endpoint.
+> **WARNING: Never expose Graph Node's administrative ports to the public**.
+>
+> - Exposing Graph Node's internal ports can lead to a full system compromise.
+> - These ports must remain **private**: JSON-RPC Admin endpoint, Indexing Status API, and PostgreSQL.
+> - Only expose 8000 (GraphQL HTTP) and 8001 (GraphQL WebSocket) if needed. Lock everything else down with firewalls or private networks.
```
encalypto, line 92: Can we use stronger language here to indicate that no ports should be exposed, even 8000 / 8001? Ideally these are proxied via `indexer-agent` instead, and that uses some kind of production-ready proxy instead

idalithb, line 92:
```suggestion
> - Do not expose 8000 (GraphQL HTTP) and 8001 (GraphQL WebSocket) directly to the internet. Even though these are used for GraphQL queries, they should ideally be proxied though `indexer-agent` and served behind a production-grade proxy. 
> - Lock everything else down with firewalls or private networks.
```

PR #975, opened 2025-06-05
Supported networks table

hayderkg:
Implements logic to display Check vs double check icons in `NetworksTable` based on service support levels from network registry

image

### DIFF
--- a/website/src/pages/en/index.json
+++ b/website/src/pages/en/index.json
@@ -88,7 +88,12 @@
       },
       "tokenApi": {
         "supported": "All endpoints supported"
-      }
+      },
+      "icons": {
+        "checkmark": "Checkmark",
+        "checkmarks": "Checkmarks"
+      },
+      "legendTitle": "Table Legend"
     }
   },
   "networkGuides": {
--- a/website/src/pages/en/index.json
+++ b/website/src/pages/en/index.json
@@ -72,6 +72,23 @@
       "substreams": "Substreams",
       "firehose": "Firehose",
       "tokenApi": "Token API"
+    },
+    "tableLegend": {
+      "subgraphs": {
+        "basic": "Subgraph Studio (No issuance)",
+        "full": "The Graph Network (Issuance)"
+      },
+      "substreams": {
+        "basic": "Base",
+        "full": "Extended (EVM Only)"
+      },
+      "firehose": {
+        "basic": "Base",
+        "full": "Extended (EVM Only)"
+      },
+      "tokenApi": {
+        "supported": "All endpoints supported"
+      }
     }
   },
   "networkGuides": {
--- a/website/src/supportedNetworks/NetworksTable.tsx
+++ b/website/src/supportedNetworks/NetworksTable.tsx
@@ -59,45 +59,45 @@ export function NetworksTable({ networks }: { networks: SupportedNetwork[] }) {
       </Callout>
 
       <div className="mb-6 overflow-clip rounded-8 border border-space-1500 bg-space-1800">
-        <div className="grid grid-cols-2 gap-px text-space-500">
+        <div className="grid grid-cols-1 gap-px text-space-500 xs:grid-cols-2">
           <div className="border-b border-r border-space-1500 p-4">
             <Text.C10 className="mb-2 uppercase text-white">Subgraphs</Text.C10>
             <div className="flex items-center gap-2">
               <Check size={4} alt="Checkmark" />
-              <span className="text-14">Subgraph Studio (No issuance)</span>
+              <span className="text-14">{t('index.supportedNetworks.tableLegend.subgraphs.basic')}</span>
             </div>
             <div className="flex items-center gap-2">
               <Checks size={4} alt="Checkmarks" />
-              <span className="text-14">The Graph Network (Issuance)</span>
+              <span className="text-14">{t('index.supportedNetworks.tableLegend.subgraphs.full')}</span>
             </div>
           </div>
           <div className="border-b border-r border-space-1500 p-4 lg:border-r-0">
             <Text.C10 className="mb-2 uppercase text-white">Substreams</Text.C10>
             <div className="flex items-center gap-2">
               <Check size={4} alt="Checkmark" />
-              <span className="text-14">Base</span>
+              <span className="text-14">{t('index.supportedNetworks.tableLegend.substreams.basic')}</span>
             </div>
             <div className="flex items-center gap-2">
               <Checks size={4} alt="Checkmarks" />
-              <span className="text-14">Extended (EVM Only)</span>
+              <span className="text-14">{t('index.supportedNetworks.tableLegend.substreams.full')}</span>
             </div>
           </div>
           <div className="border-b border-r border-space-1500 p-4">
             <Text.C10 className="mb-2 uppercase text-white">Firehose</Text.C10>
             <div className="flex items-center gap-2">
               <Check size={4} alt="Checkmark" />
-              <span className="text-14">Base</span>
+              <span className="text-14">{t('index.supportedNetworks.tableLegend.firehose.basic')}</span>
             </div>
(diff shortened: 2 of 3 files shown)


hayderkg:
> @hayderkg – Can you please add a screenshot to your PR? 🙏

Done! 🙌

website/src/supportedNetworks/NetworksTable.tsx
```diff
@@ -58,6 +58,51 @@ export function NetworksTable({ networks }: { networks: SupportedNetwork[] }) {
         </p>
       </Callout>
 
+      <div className="mb-6 overflow-clip rounded-8 border border-space-1500 bg-space-1800">
```
benface, line 61: Should we make it clearer to screen readers that this is a legend? I'm thinking this element could be an `aside` with a label (either with an `sr-only` heading inside it, or probably better, an `aria-labelledby` pointing to a `hidden` element). What do you think?

website/src/supportedNetworks/NetworksTable.tsx
```diff
@@ -58,6 +58,51 @@ export function NetworksTable({ networks }: { networks: SupportedNetwork[] }) {
...
+        <div className="grid grid-cols-1 gap-px text-space-500 xs:grid-cols-2">
+          <div className="border-b border-r border-space-1500 p-4">
+            <Text.C10 className="mb-2 uppercase text-white">Subgraphs</Text.C10>
+            <div className="flex items-center gap-2">
+              <Check size={4} alt="Checkmark" />
```
benface, line 66: I'm noticing all the icons in this legend have the same `size`, so lil' GDS trick: you can omit the `size` prop and use `nested-icon:default-size-4` on the container. Of course, it's also totally fine to use the prop, just a bit repetitive (but we already have necessary repetition with the `alt` I guess, unless we wrap `Check` into a local component... probably not worth it).

website/src/supportedNetworks/NetworksTable.tsx
```diff
@@ -58,6 +58,51 @@ export function NetworksTable({ networks }: { networks: SupportedNetwork[] }) {
...
+              <span className="text-14">{t('index.supportedNetworks.tableLegend.subgraphs.basic')}</span>
+            </div>
+            <div className="flex items-center gap-2">
+              <Checks size={4} alt="Checkmarks" />
+              <span className="text-14">{t('index.supportedNetworks.tableLegend.subgraphs.full')}</span>
+            </div>
+          </div>
+          <div className="border-b border-r border-space-1500 p-4 lg:border-r-0">
+            <Text.C10 className="mb-2 uppercase text-white">Substreams</Text.C10>
```
benface, line 75: I should have marked `Text` as `@deprecated` a while ago, my bad... but the new GDS way is the `text-c10` class.

website/src/supportedNetworks/NetworksTable.tsx
benface: Also, should the checks be centered with all the lines when the label wraps, or just with the first line?

CleanShot 2025-06-05 at 15 40 41@2x

I think just the first line would look better; you can achieve that by removing `items-center` from the `flex` containers and adding `h-[1lh]` on the icons.

website/src/supportedNetworks/NetworksTable.tsx
```diff
@@ -58,6 +58,51 @@ export function NetworksTable({ networks }: { networks: SupportedNetwork[] }) {
         </p>
       </Callout>
 
+      <div className="mb-6 overflow-clip rounded-8 border border-space-1500 bg-space-1800">
+        <div className="grid grid-cols-1 gap-px text-space-500 xs:grid-cols-2">
+          <div className="border-b border-r border-space-1500 p-4">
+            <Text.C10 className="mb-2 uppercase text-white">Subgraphs</Text.C10>
```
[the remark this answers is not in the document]
benface: My bad; like you said on Slack, these are product names so no need to translate them. Just "Checkmark" / "Checkmarks".

website/src/supportedNetworks/NetworksTable.tsx
hayderkg: wow, I think it's great!

website/src/supportedNetworks/NetworksTable.tsx
hayderkg: Amazing, the only thing is that `1lh` is taller than the line-height of `text-14`, I've added `0.75lh` which is the same

website/src/supportedNetworks/utils.ts
```diff
@@ -41,6 +41,29 @@ export const getIconVariant = (networkId: string): 'mono' | 'branded' => {
...
+  if (substreamCount >= 2) return 'full'
+  return 'basic'
+}
+
+export const getFirehoseSupportLevel = (network: any): 'none' | 'basic' | 'full' => {
+  const firehoseCount = network.services.firehose?.length || 0
+  if (firehoseCount === 0) return 'none'
+  if (firehoseCount >= 2) return 'full'
+  return 'basic'
+}
```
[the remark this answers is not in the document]
hayderkg: whoops! thanks

website/src/pages/en/index.json
```diff
@@ -72,6 +72,28 @@
...
+        "full": "Extended (EVM Only)"
+      },
+      "tokenApi": {
+        "supported": "All endpoints supported"
+      },
+      "icons": {
+        "checkmark": "Checkmark",
+        "checkmarks": "Checkmarks"
+      },
+      "legendTitle": "Table Legend"
```
benface, line 96: 🙌

website/src/supportedNetworks/utils.ts
```diff
@@ -41,6 +41,29 @@ export const getIconVariant = (networkId: string): 'mono' | 'branded' => {
...
+  if (substreamCount >= 2) return 'full'
+  return 'basic'
+}
+
+export const getFirehoseSupportLevel = (network: Network): 'none' | 'basic' | 'full' => {
+  const firehoseCount = network.services.firehose?.length || 0
+  if (firehoseCount === 0) return 'none'
+  if (firehoseCount >= 2) return 'full'
+  return 'basic'
+}
```
benface, line 65:
```suggestion
export const getSubgraphsSupportLevel = (network: Network) => {
  const hasSubgraphs = Boolean(network.services.subgraphs?.length || network.services.sps?.length)
  if (!hasSubgraphs) return 'none'
  if (network.issuanceRewards) return 'full'
  return 'basic'
}

export const getSubstreamsSupportLevel = (network: Network) => {
  const substreamCount = network.services.substreams?.length || 0
  if (substreamCount === 0) return 'none'
  if (substreamCount >= 2) return 'full'
  return 'basic'
}

export const getFirehoseSupportLevel = (network: Network) => {
  const firehoseCount = network.services.firehose?.length || 0
  if (firehoseCount === 0) return 'none'
  if (firehoseCount >= 2) return 'full'
  return 'basic'
}
```
Or do you feel that the return types are helpful? Don't want to insist, I know there's some subjectivity on that topic... but I truly don't see the point; if someone accidentally changes the type that's returned in the future, I feel like TS will catch it when we actually use the value (e.g. `if (network.subgraphsSupportLevel === 'none') will fail if `subgraphsSupportLevel` can never be `none`). I definitely have a bias for "less code is better", but I swear there are cases where I find return types helpful; I just fail to see this as one of them.

website/src/supportedNetworks/NetworksTable.tsx
benface: Oh, but the whole point of using `1lh` is to ensure they're aligned regardless of the font size / line height, which can change (e.g. a GDS update could change it and break this). Instead, you should put the `text-14` on the parent `flex`. I don't see the benefit of `0.75lh` over specifying a height in pixels.

benface (approved):

PR #980, opened 2025-06-10
Style inline images properly

benface:
### Before

CleanShot 2025-06-10 at 12 07 42@2x

### After

CleanShot 2025-06-10 at 12 00 37@2x


0xa3k5 (approved):
perfect

PR #992, opened 2025-06-26
Auto-expand/collapse nav items + some other improvements

benface:
This PR squashes a few TODOs that existed in the code, notably these:

> TODOs about the navigation system:
> - Auto-expand the current item's ancestors on load and when the route changes
> - Remember which groups/items are expanded at every level, so that closing and re-expanding an ancestor doesn't affect its children's state
> - Remember if groups/items were manually expanded (with the caret) or auto-expanded (either by clicking on its title or by navigating to it or one of its children), then auto-close those that were auto-expanded when the user navigates to a different group/item

It also updates dependencies (notably Prettier, hence the large number of files changed – sorry), tweaks the `Callout` styles (thanks @0xa3k5), replaces ✅ emojis with GDS (phosphor) check icons (thanks @hayderkg), and colocates our custom MDX (rehype and remark) plugins in the `website` package, as they were already tightly coupled to it (I converted them to TS while I was at it).


website/src/components/Callout.tsx
```diff
@@ -53,7 +53,7 @@ export const Callout = ({
       ])}
       {...(props as ComponentProps<'div'>)}
     >
-      <div className="flex size-6 shrink-0 items-center justify-center">
+      <div className="flex h-[1lh] shrink-0 items-center justify-center">
```
0xa3k5, line 56: 🤩

website/src/components/Image.tsx
```diff
@@ -1,15 +1,10 @@
...
 import { classNames } from '@edgeandnode/gds'
 
-export interface ImageProps extends Omit<ComponentProps<'img'>, 'src'> {
-  src?: ComponentProps<typeof NextImage>['src']
-}
+interface ImageProps extends ComponentProps<'img'> {}
 
-export const Image = ({ src: passedSrc, alt, className, ...props }: ImageProps) => {
-  const src =
-    typeof passedSrc === 'object' ? ('default' in passedSrc ? passedSrc.default.src : passedSrc.src) : passedSrc
```
0xa3k5, line 12: this was not needed?

website/src/components/TimeIcon.tsx
```diff
@@ -5,7 +5,7 @@ import { HourglassDynamic } from '@edgeandnode/gds/icons'
 
 import { useI18n } from '@/i18n'
 
-export interface TimeIconProps extends ComponentProps<'div'> {
+interface TimeIconProps extends ComponentProps<'div'> {
```
0xa3k5, line 8: ah. nice catch it was probably my mistake 😅

0xa3k5 (approved):
❤️

website/src/components/TimeIcon.tsx
benface: Not really a mistake, haha, just removed it for consistency.

PR #995, opened 2025-07-02
Redirect Graph Assistant page

hayderkg:


kmjones1979 (commented):
lgtm

benface (changes requested):
This won't work on production (or staging) because the docs are a static build at the moment (no Next server). We need to add the redirect to `nginx.conf`.

benface (approved):
I fixed it. :)

PR #1000, opened 2025-07-08
Update utils.ts

UrsaPolarisRecords:
REVIEW REQUIRED. 
Removed hard-coded Token API networks, added what is hopefully an automatic fetch for Token API networks.

PLEASE REVIEW THIS PR. It was implemented with advice from Claude. Not sure where these parameters are used, so I am unable to check if it works. Happy to let docs team make this update instead, but on the off chance this is correct feel free to merge.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -118,8 +118,8 @@ importers:
         specifier: ^7.3.0
         version: 7.3.0
       '@pinax/graph-networks-registry':
-        specifier: ^0.6.7
-        version: 0.6.7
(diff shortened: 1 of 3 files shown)


UrsaPolarisRecords:
Yay for PR number 1000! 🎉

UrsaPolarisRecords:
Hmm, failing typecheck. Claude seems to want me to make all sorts of edits, but I think in this case it is wise to let the docs team take over

idalithb (approved):

PR #1038, opened 2025-11-06
Dex Section

idalithb:
@benface couldn't find the markdown for accordions inside pages. I know we have a component for the nav, but I haven't applied it inside pages.


benface (approved):
🙏

PR #1051, opened 2025-11-19
Update Token API MCP documentation

0237h:
**introduction.mdx**
- Adds the two types of MCP
- Emphasizes REST easier to use for LLMs than SQL 
- Misc editing

**claude.mdx**
- Adds REST MCP config
- Updates config commands
- Adds depreciation notice for `--sse-url` option
- Improves troubleshooting section

**cline.mdx**
Same as Claude

**cursor.mdx**
Same as Claude

**faq.mdx**
- Adds `How do I choose between the REST-MCP and the SQL-MCP and which one should I use ?`
- Adds `Will using the REST-MCP counts towards my Token API usage ?`

### DIFF
--- a/website/src/pages/en/ai-suite/token-api-mcp/claude.mdx
+++ b/website/src/pages/en/ai-suite/token-api-mcp/claude.mdx
@@ -19,6 +19,7 @@ Create or edit your `claude_desktop_config.json` file.
 - Linux: `.config/Claude/claude_desktop_config.json`
 
 **REST-based MCP**
+
 ```json label="claude_desktop_config.json"
 {
   "mcpServers": {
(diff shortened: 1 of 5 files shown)


matthewdarwin:
We should definately suggest to users to use the REST MCP as the SQL one is easy to get things wrong and generate inefffient queries.   By default use that one unless there is some special need to use SQL.

In this PR we won't address pricing for SQL MCP, other than you need to be on PRO or ENTERPRISE plan.  Pricing page can be updated in a future PR.

benface (commented):
Found some typos, but otherwise LGTM.

benface (approved):
🙌

PR #1054, opened 2025-11-27
Token API docs optimization

nsjames:
These changes aim to reduce onboarding friction for the Token API.

I've added a `CodeTabs` and `CodeTab` component that can be used to add multi-language code blocks as well, styling might not align with branding so let me know if it needs changes.

image

I've also moved some routes around: 
- Added a rewrite for `endpoint-pricing`
- Moved `endpoint-pricing` to `/endpoints/pricing`
- Moved support matrix section from quickstart to `endpoints/support`

It seems like moving that route with the `move-pages` script has fixed some of the misaligned `_meta.json` files in other languages, but I don't have enough historical context about the progression of this repository to know if those were changed in the forward direction or if there is a regression there.


benface (approved):
Thank you so much!

PR #1066, opened 2026-01-06
Removed upgrade to v1 pages and settings

jhjhjh94:
Removed upgrade to v1 pages and settings


benface (changes requested):
@jhjhjh94 – LGTM but there is a formatting issue, can you please run `pnpm check:fix`?

PR #1098, opened 2026-04-28
Change Sei Docs Link

itsjerryokolo:
## Summary
- Update the rendered Sei mainnet docs link to https://docs.sei.io
- Keep the override scoped to the Sei supported-network entry

## Validation
- pnpm install --frozen-lockfile
- pnpm build
- Verified the built Sei page contains docs.sei.io

### DIFF
--- a/website/src/supportedNetworks/utils.ts
+++ b/website/src/supportedNetworks/utils.ts
@@ -44,6 +44,7 @@ export async function getSupportedNetworks() {
       return [
         {
           ...network,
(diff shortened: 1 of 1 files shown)


itsjerryokolo:
This change was requested by the Sei Foundation

benface (changes requested):
Thanks @itsjerryokolo, but the change should be made upstream in https://github.com/graphprotocol/networks-registry/blob/main/registry/eip155/sei-mainnet.json and then there would be no need for an override.

itsjerryokolo:
Thanks @benface 
I'll make the change there.

PR #1110, opened 2026-06-15
Token API redirect patch

brandonleekramer:
The URL redirect on line 243 of the nginx.conf file was wrong, triggering a 404 on the former pricing page.


benface:
Superseded by #1111

